Issue #19, opened 2017-07-01
Planning meeting - July 2 at 13:00


rebeccaradding:
@hibaakel @RQ16 @Suuuuha @Karyum @MohamedOmarii @HilbertSpitzer 

Since tomorrow (Sunday) is the only day you will all be in attendance, I suggest we meet at lunch to discuss the plan for the Open Tourism Platform. 

@sofer will explain:
- Overall plan
- Individual assignments
- Schedule
- Budget
- Invoicing

For more, visit this document: https://docs.google.com/document/d/1vt7LdZOgEqx6nuxWDCnvDHL8qVWaln4MkHGdTglKcX4/edit?usp=sharing

mattlub:
closed as meeting minutes added

PR #22, opened 2017-07-03
add minutes from group meeting on 2 July

m4v15:
relates #19


mattlub (approved):

mattlub (changes requested):
probably put it in a folder or something?

m4v15:
Good point, done

Issue #30, opened 2017-07-05
should we have a 'places' table in the db?


mattlub:
- I know this may not be required for the initial 4 apps.
- But I can see it being important for tourism.
- e.g. the Basilica would not fit anywhere in the currently proposed data model.

Thoughts?

sofer:
Also, Hospitals, schools, libraries...

If you go for both "places" and "businesses", you might have to introduce another relationship in the schema ("a business has_many locations"). Not sure if that is necessary. Perhaps it is just a question of naming conventions.

mantagen:
Could be useful 
https://docs.mongodb.com/manual/geospatial-queries/

des-des:
@sofer right now we are allowing an organisation / business to sign up as a user, this leaves us space to to grow out the idea of having orgs in the future, without having to deal with all that complexity up front

mattlub:
closing as initial specs now merged.

Issue #15, opened 2017-06-29
add permissions information to API documentation


mattlub:
- add specific permissions info for relevant endpoints.
- also maybe add to permissions section, a summary of each of the different permissions options

m4v15:
Potentially get v0 going without permissions, so this could be made priority 2 or something?

mattlub:
either
- without permissions, or
- with a single admin role which has permission to create/update/delete

@des-des @m4v15 ?

des-des:
@mattlub @m4v15 For the initial version, we have
  1. guest (unauthenticated)
  2. member
  3. admin

There are some ownership problems around the artisan products, but I think we should ignore these for now

mattlub:
@des-des what permissions did you have in mind for member and admin?
the way I saw it, after talking to Dan:

1. guest, unauthenticated, or basic member/user (can view businesses, events, users, add businesses)
2. business owner (can edit business they own, add events)
3. admin e.g. one of the volunteers adding businesses (can edit everything, including assign owner to business)

I'm looking forward to discussing this.
(also thoughts @m4v15?)

m4v15:
Yeah, pretty much, except I think maybe we said guest's can't directly add business, but instead will just fill a form that sends an email to an admin (us?), who can then verify and add if legit?

m4v15:
I'm thinking something like:

### Permissions Table
||Users|Businesses|Events|Products|
---|---|---|---|---
GET|Admin Only|Open|Open|Open|
POST|Open? Anyone can reg, defaults guest|Admin? Anyone can submit to add a business|Business owners only| Business Owners only|
PUT|Admin Only|Admin, The Specific Business Owner|Admin, Event Creator (which would be the same as the business owner)|Admin, Product Creator (again, same as BO)|
DELETE|Admin Only|Admin, The Specific Business Owner|Admin, Event Creator (which would be the same as the business owner)|Admin, Product Creator (again, same as BO)|

m4v15:
So Users would be role based: Admin, Biz Owner, Guest.

Then, business would all have an owner field - PUT and DELETE requests would then check the users JWT to see if they are a) *A* business owner and b) that user's ID matches the id in the businesses owner field. If either fail, denied.

Similar for Events/Products PUT and DELETE. They would have a field with the ID of the business that created them. Then you can check that the user is the owner of the Business that created them.

[...]

Issue #13, opened 2017-06-29
document (provisional at least) db schema


mattlub:
for the 4 tables:
- user
- business
- event
- product

related #10

m4v15:
|User| |
|---|---|
|id| integer, Unique|
|username| String, Unique|
|password| String, hashed|
|email| String|
|role| String|

m4v15:
|Business| |
|---|---|
|id| integer, Unique Key|
|ownerId| integer, foreign Key|
|name| String|
|address| String|
|category| String|
|accessOptions|String|
|openingHours|String|
|description|String|
|website|String|
|phone|Integer|
|email|String|
|imageLink|String|
|lat|float|
|lng|float|

m4v15:
|Event| |
|---|---|
|id| integer, Unique|
|bizId| integer, foreignKey|
|name| String|
|location| String (default to Business)|
|category| String (default to Business)|
|accessOptions|String|
|startTime|String|
|endTime|String|
|date| Date|
|description|String |
|cost| String|
|imageLink|String (default to Business)|

m4v15:
|Product| |
|---|---|
|id| integer, Unique|
|bizId| integer, foreignKey|
|name| String|
|type| String|
|description|String |
|cost| String|
|imageLink|String (default to Business)|
|numberAvaliable| integer|

mattlub:
looks pretty good, put these into a PR so we can start to get some stuff merged?

mattlub:
closing as inital spec merged

Issue #12, opened 2017-06-29
create API endpoint documentation


mattlub:
- CRUD endpoints for each of:
  - user
  - business (is this even the definite name?)
  - event 
  - product
  - invite
  - signup
- decide on good format: e.g.
  - github is quite good imo: https://developer.github.com/v3/repos/
  - https://bocoup.com/blog/documenting-your-api

mattlub:
## Users
- create user
- show all users
- show specific user
- update user
- delete user

----

### Show specific user
Returns json for a specific user.
```
GET /users/:id
```

- #### URL Params (required)
`id=[integer]`

- #### Query Params
None

- #### Required Permissions
canViewUsers

- #### Success Response
**Code:** ```200```
**Content:** ```{ id : 12, name : "Michael Bloom" }```
 
- #### Error Response
**Code:** ```404 NOT FOUND```
**Content:** ```{ error : "User doesn't exist" }```

OR

**Code:** ```401 UNAUTHORIZED```
**Content:** ```{ error : "You are unauthorized to make this request." }```

- #### Sample Call

```javascript
fetch('/users/1')
```

mattlub:
@des-des thoughts on the formatting above?
We should probably be sure about the formatting before writing out like 20 endpoints right?

des-des:
@mattlub prob a table?

mattlub:
closed as docs have been merged #90

PR #31, opened 2017-07-05
create data spec

mattlub:
the specs file, this is a precursor to both the crud api and the database schema.

related #4 #12 #13 #15 #30


m4v15:
Is it worth having this as part of the readme?

eliasmalik:
I haven't been part of the conversations on this, so I'm not particularly well placed to review the appropriateness of the data model, but I would say that the use of `SUPER` and `OWNER` throughout is confusing to me.

In [user roles](https://github.com/foundersandcoders/open-tourism-platform/blob/7e6d5545ea6ae5290acbb749016b089e77ded950/data_spec_v1.md#user-roles):
```
* SUPER - OWNER of the open platform.
* ADMIN - ...
* BASIC - ...
```
This implies that there are three roles: `SUPER`, `ADMIN` and `BASIC`. But later, both `OWNER` _and_ `SUPER` are used as a roles. For example: `permissions: only OWNER or SUPER`. I don't know what this means.

Also I think there's a typo: should be `shekels`, not `shekles`.

m4v15:
@eliascodes I think what @des-des means by OWNER in the tables is the owner of that particular user/place/product/event:
> BASIC - can have ownership of organisations / their data / events / products

Places/products/events don't have to have owners (they can be assigned by admins/super) but if they do, then that owner can alter the fields.

des-des:
@eliascodes yeah that was super unclear, tried to add some clarifaction

eliasmalik:
@des-des @m4v15 thanks, that's better. No more comments 👍

mattlub:
1. 'API permissions' in Content section is now defunct
2. for User CREATE permissions I would have role=BASIC:**any**, role=ADMIN:SUPER, no reason to stop someone creating a user?

data_spec_v1.md
```diff
@@ -0,0 +1,164 @@
...
+#### Fields
+
+
+name | type | required | permissions
+---|---|---|---
+id | KEY | true |-
+createdAt| DATE | - | -
+updatedAt| DATE | - | -
+isPublic | BOOL | true | only OWNER or SUPER (not ADMIN?)
+username (do we need this?) | TEXT | false | only OWNER or SUPER
```
m4v15: I guess there's a fairly likely future feature that will be something along the lines of reviews or a CMS type thing. In which case we will want authors for the content, i.e. usernames? Or that could be done manually with each post or whatever.

Using mongo if we start off without and just log in using email and password, adding usernames if needed in the future shouldn't be too hard I guess? So maybe no, we don't need them.

data_spec_v1.md
```diff
@@ -0,0 +1,164 @@
...
+password | HASH | true | only OWNER or SUPER
+name| STRING | true | only OWNER or SUPER
+organisationName | TEXT | false | only OWNER or SUPER
+organisationDescription | TEXT | false | only OWNER or SUPER
```
m4v15: Is this necessary? Should places not have a description, rather than organisations?

The place Liwan Cafe would need a description, but not the "organisation" Liwan. Or maybe both I guess. I think places definitely should have though though.

data_spec_v1.md
```diff
@@ -0,0 +1,164 @@
...
+id | KEY | true |-
+createdAt| DATE | - | -
+updatedAt| DATE | - | -
+ownerId | FOREIGN KEY | false | ADMIN
+name| STRING | | OWNER
+address | ADDRESS | true | OWNER
+location| LOCATION | | OWNER
+isVerified | BOOL | true | ADMIN
+category| ?? do we have a categories table ? / is this an enum| | OWNER
+accessibilityOptions| ? | true | OWNER
```
m4v15: Only editable by admins for verification purposes? Or not for v0.

data_spec_v1.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+name | type | required | permissions
+---|---|---|---
+id | KEY | true |-
+createdAt| DATE | - | -
+updatedAt| DATE | - | -
+ownerId | FOREIGN KEY | false | ADMIN
+name| STRING | | OWNER
+location| LOCATION | | OWNER
```
m4v15, line 121: Is this a foreign key referencing the places table?

data_spec_v1.md
```diff
@@ -0,0 +1,164 @@
...
+id | KEY | true |-
+createdAt| DATE | - | -
+updatedAt| DATE | - | -
+ownerId | FOREIGN KEY | false | ADMIN
+name| STRING | | OWNER
+category| ?? do we have a categories table ? / is this an enum| | OWNER
+description| TEXT | | OWNER
+cost (shekles) | DOUBLE | | OWNER
+imageUrl| TEXT | | OWNER
+location? | LOCATION | | OWNER
```
m4v15: Probably don't need location for products?

mattlub:
remove reference to 'moderator' from intro section and permissions section, what was 'moderator' is  ADMIN now

data_spec_v1.md
des-des: Not for v1

data_spec_v1.md
mattlub: kept it here, added it to place as well

data_spec_v1.md
mattlub: I presume location is a data type

mattlub:
@des-des @m4v15 made a commit with minor updates.

data_spec_v1.md
m4v15: I meant more should this field be a foreign key that references the places table? @des-des

m4v15 (approved):

des-des (approved):

PR #37, opened 2017-07-09
Set up basic server

mattlub:
related #34 

I also added standard linter, nodemon, pre-commit hook.


m4v15 (approved):

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "open-tourism-platform",
+  "version": "1.0.0",
+  "description": "An open data platform for Nazareth tourism",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "node src/start.js",
+    "lint": "./node_modules/.bin/standard src/*.js",
```
des-des: Are you sure
`"lint": "standard src/*.js"` does not work

des-des: Are you sure you need that pattern also: `src/*.js`

My experience with standard is that it just works..

package.json
```diff
@@ -0,0 +1,29 @@
...
+    "dev": "./node_modules/.bin/nodemon src/start.js"
```
des-des: `nodemon src/start.js`

des-des (changes requested):

des-des (approved):

Issue #36, opened 2017-07-09
deploy to heroku


mattlub:
relates #32

mattlub:
I've put it on heroku with a mongolab db, https://dashboard.heroku.com/apps/nazareth-open-tourism-platform

m4v15:
Can you give me permission to view plz

mattlub:
closed as it's now LIVE

mattlub:
@des-des @mantagen @m4v15  @sofer it's on my heroku account right now, I can give you all collaborator status or we can transfer it elsewhere at some point.

des-des:
@mattlub That great for now, we will work out what to do with it next sprint

Issue #35, opened 2017-07-09
set up local mongo and link to server 


mattlub:
related #32

mattlub:
@m4v15 use mongoose from the start?

PR #38, opened 2017-07-09
Set up mongo

mattlub:
adds mongoose and mongo, and a db connect file, ready for deployment to heroku

relates #35 #36 

builds on #37

### DIFF
--- a/db/connect.js
+++ b/db/connect.js
@@ -1,0 +1,12 @@
+const mongoose = require('mongoose')
+
+// http://mongoosejs.com/docs/promises.html
+mongoose.Promise = global.Promise
+
+const dbUrl = 'mongodb://localhost/testdb'
+
+mongoose.connect(dbUrl, {
+  useMongoClient: true
+})
+
+module.exports = mongoose.connection
--- a/db/connect.js
+++ b/db/connect.js
@@ -3,9 +3,9 @@ const mongoose = require('mongoose')
 // http://mongoosejs.com/docs/promises.html
 mongoose.Promise = global.Promise
 
-const dbUrl = 'mongodb://localhost/testdb'
+const dbURI = process.env.MONGODB_URI
 
-mongoose.connect(dbUrl, {
+mongoose.connect(dbURI, {
   useMongoClient: true
 })
 
@@ -13,7 +13,7 @@ mongoose.connection.on('error', err => {
   console.log('connection error: ' + err)
 })
 
-mongoose.connection.on('disconnected', () => {  
+mongoose.connection.on('disconnected', () => {
   console.log('Mongoose default connection disconnected.')
 })
 
--- a/db/connect.js
+++ b/db/connect.js
@@ -9,4 +9,12 @@ mongoose.connect(dbUrl, {
   useMongoClient: true
 })
 
+mongoose.connection.on('error', err => {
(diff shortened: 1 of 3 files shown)


db/connect.js
```diff
@@ -0,0 +1,12 @@
+const mongoose = require('mongoose')
+
+// http://mongoosejs.com/docs/promises.html
+mongoose.Promise = global.Promise
+
+const dbUrl = 'mongodb://localhost/testdb'
```
m4v15: Do we want to set up an process.env.PRODUCTION variable now to connect to different databases?

Or shall we just do this for now.

I think maybe just do this for now actually since it's not hard to put things in an IF statement

src/start.js
```diff
@@ -0,0 +1,26 @@
+const mongoose = require('mongoose')
+const server = require('./server.js')
+
+// connect to the db
+require('../db/connect.js')
+const db = mongoose.connection
+
+const port = process.env.PORT || 3000
+
+db.on('error', err => console.log('connection error: ' + err))
```
m4v15: I think it's better to use:
```js
db.on('error', console.error.bind(console, 'connection error:'));
```

Tbh, I don't really know why and have researched it before, but it's in the [docs for mongoose](http://mongoosejs.com/docs/).

https://www.tjvantoll.com/2015/12/29/console-error-bind/

m4v15 (changes requested):

db/connect.js
mattlub: yeh- think we can do it when we hook it up with heroku

m4v15 (approved):

db/connect.js
```diff
@@ -0,0 +1,20 @@
+const mongoose = require('mongoose')
+
+// http://mongoosejs.com/docs/promises.html
+mongoose.Promise = global.Promise
```
m4v15, line 4: I'm not like, 100% fluent in promises or whatever, but I understand it might be good to use bluebird?

> While mpromise is sufficient for basic use cases, advanced users may want to plug in their favorite ES6-style promises library like bluebird, or just use native ES6 promises. Just set mongoose.Promise to your favorite ES6-style promise constructor and mongoose will use it.

Although are we jsut gonna be a basic use case

m4v15: I'll research promises now tbh

des-des: @m4v15 I use native promises, I think especially if you do not have experience with promises, native is totes fine

db/connect.js
```diff
@@ -0,0 +1,20 @@
...
+
+const dbURI = process.env.MONGODB_URI
+
+mongoose.connect(dbURI, {
+  useMongoClient: true
```
des-des, line 9: :+1:

db/connect.js
m4v15: 👍

src/start.js
```diff
@@ -1,10 +1,19 @@
...
 
-server.listen(port, err => {
-  if (err) {
-    throw err
-  }
-  console.log(`Server listening on port ${port}`)
+// connect to the db
+const dbConnection = require('../db/connect.js')
+
+dbConnection.once('open', () => {
```
des-des, line 11: Wikid good good job here

des-des (approved):

PR #43, opened 2017-07-10
create user table schema and model

m4v15:
relates #39

### DIFF
--- a/db/models/User.js
+++ b/db/models/User.js
@@ -1,0 +1,22 @@
+const mongoose = require('mongoose')
+
+const roles = ['BASIC', 'ADMIN', 'SUPER']
+
+const userSchema = mongoose.Schema(
+  {
+    isPublic: Boolean,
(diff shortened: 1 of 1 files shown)


db/models/User.js
```diff
@@ -0,0 +1,22 @@
...
+
+const roles = ['BASIC', 'ADMIN', 'SUPER']
+
+const userSchema = mongoose.Schema(
+  {
+    isPublic: Boolean,
+    username: { type: String, required: true, index: { unique: true } },
+    password: { type: String, required: true },
+    name: { type: String, required: true },
+    organisationName: { type: String, required: false },
```
mattlub: required: false is default, doesn't need to be specified

mattlub (changes requested):

mattlub (approved):

des-des (approved):

PR #45, opened 2017-07-10
create events table schema and model

m4v15:
relates #41

### DIFF
--- a/db/models/Events.js
+++ b/db/models/Events.js
@@ -1,11 +1,11 @@
 const mongoose = require('mongoose')
 
 const categories = ['music', 'dining']
-const accessibilityOptions = ['braille menu', 'wheelchair-friendly']
+const accessibilityOptions = ['braille-menu', 'wheelchair-friendly']
 
 const eventSchema = mongoose.Schema(
   {
-    ownerId: {type: mongoose.Schema.Types.ObjectId, ref: 'User'}
+    ownerId: {type: mongoose.Schema.Types.ObjectId, ref: 'User'},
     name: { type: String, required: true },
     category: { type: [String], enum: categories, required: true },
     location: {type: mongoose.Schema.Types.ObjectId, ref: 'Place'},
@@ -13,7 +13,7 @@ const eventSchema = mongoose.Schema(
     startTime: Date,
     endTime: Date,
     description: String,
-    cost: String
+    cost: String,
     imageUrl: String
   },
   {
--- a/db/models/Events.js
+++ b/db/models/Events.js
@@ -5,10 +5,10 @@ const accessibilityOptions = ['braille-menu', 'wheelchair-friendly']
 
 const eventSchema = mongoose.Schema(
   {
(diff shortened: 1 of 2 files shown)


db/models/Events.js
```diff
@@ -0,0 +1,24 @@
+const mongoose = require('mongoose')
+
+const categories = ['music', 'dining']
+const accessibilityOptions = ['braille-menu', 'wheelchair-friendly']
+
+const eventSchema = mongoose.Schema(
+  {
+    ownerId: {type: mongoose.Schema.Types.ObjectId, ref: 'User'},
```
mattlub: spacing maves :(

mattlub (changes requested):

mattlub (approved):

db/models/Events.js
```diff
@@ -0,0 +1,24 @@
...
+const categories = ['music', 'dining']
+const accessibilityOptions = ['braille-menu', 'wheelchair-friendly']
+
+const eventSchema = mongoose.Schema(
+  {
+    ownerId: { type: mongoose.Schema.Types.ObjectId, ref: 'User' },
+    name: { type: String, required: true },
+    category: { type: [String], enum: categories, required: true },
+    location: { type: mongoose.Schema.Types.ObjectId, ref: 'Place' },
+    accessibilityOptions: { type: [String], enum: accessibilityOptions, required: false },
```
des-des: An enum can only hold one value at once..

db/models/Events.js
```diff
@@ -0,0 +1,24 @@
+const mongoose = require('mongoose')
+
+const categories = ['music', 'dining']
+const accessibilityOptions = ['braille-menu', 'wheelchair-friendly']
+
+const eventSchema = mongoose.Schema(
+  {
+    ownerId: { type: mongoose.Schema.Types.ObjectId, ref: 'User' },
+    name: { type: String, required: true },
+    category: { type: [String], enum: categories, required: true },
```
des-des: Yeah, I may have been wrong in the schema, I expect you may want to have more than one category.

des-des (changes requested):

db/models/Events.js
mattlub: we have a [String]- an array of strings. which apparently works (https://stackoverflow.com/questions/27447876/is-it-possible-to-create-a-multi-select-enum-in-mongoose) though we haven't tested it yet

mattlub: maybe we need to test it.

mattlub: just checked- as it is doesn't work, but ```category: [{ type: String, enum: categories}]``` does work. In order to require at least one value, a custom validator would have to be implemented though, as even with ```required: true```, it would allow an empty list @m4v15

db/models/Events.js
mattlub: as above ^

des-des (approved):

des-des:
ops!

Issue #48, opened 2017-07-11
CRUD endpoints for Users table


m4v15:
with tests

relates #26

PR #53, opened 2017-07-12
Add router, users endpoints, handlers shells

mattlub:
restructuring of folders has made this look bigger than it is

related #55, #48


mattlub:
@des-des we've written some tests, what do you think?

src/routes.js
```diff
@@ -0,0 +1,12 @@
+const usersHandlers = require('./handlers/users')
+
+const router = require('express').Router()
+
+// user routes
+router.get('/users', usersHandlers.get)
```
des-des: We can go
```js
router.route('/users')
   .get(usersHandlers.getAll)
   .post(userHandlers.create)

router.route('/users/:id')
   .get(getById)
   .put
..
```
Other Thing is that they are called controllers, not handlers, you should have a folder called controllers

tests/index.test.js
```diff
@@ -1,16 +1,93 @@
...
+    .then(() => {
+      t.end()
+    })
+    .catch(err => t.end(err))
+}
 
 dbConnection.once('open', () => {
-  require('./test_1.test.js')
-  require('./test_2.test.js')
+  tape('test /users when nothing in database', (t) => {
```
des-des: Yeah, you need to move this into its own file.

Prob: `test/controllers/user.test.js`

.coveralls.yml
```diff
@@ -0,0 +1 @@
+repo_token: 0CzkQAXABFpjYwwk9l0sCx1y0hwtfGxmC
```
des-des: https://github.com/lemurheavy/coveralls-public/issues/208

You might need to generate a new token

tests/controllers/users.test.js
```diff
@@ -0,0 +1,58 @@
...
+          t.equal(res.body.length, 1, 'response body should be an array with length 1')
+          t.equal(res.body[0].username, validUser1.username, 'correct username is returned')
+          dropCollectionAndEnd(User, t)
+        })
+    })
+    .catch(err => t.end(err))
+})
+
+tape('test /users GET returns longer list', t => {
+  User.create(validUser1, validUser2)
```
des-des: I do not think you need the middle test,

am easy if you want to keep it, but its lots of extra test, and I am not sure that it gives you anything

des-des: ie Why not jump straight to 2 users

tests/controllers/users.test.js
```diff
@@ -0,0 +1,58 @@
...
+          if (err) t.fail(err)
+          // check our get path returns that user correctly
+          t.equal(res.body.length, 2, 'response body should be an array with length 1')
+          t.ok(res.body.map((user) => user.username).includes(validUser1.username), 'mattlub has been added')
+          t.ok(res.body.map((user) => user.username).includes(validUser2.username), 'm4v15 has been added')
+          dropCollectionAndEnd(User, t)
+        })
+    })
+    .catch(err => t.end(err))
+})
```
des-des, line 38: Rather than having the middle test, I would try and provoke an error.

This will become more important when you start to build out error handling + validation. But if you can, try to hit the 500.

des-des (changes requested):

tests/controllers/users.test.js
mattlub: only way to do that would be to have the test run before the database connection?

Is it necessary?

tests/controllers/users.test.js
mattlub: yeh agreed

tests/controllers/users.test.js
des-des: no thats fine

src/routes.js
```diff
@@ -0,0 +1,14 @@
+const usersControllers = require('./controllers/users')
+
+const router = require('express').Router()
+
+// user routes
+router.route('/users')
+  .get(usersControllers.get)
```
des-des: call this getAll

src/routes.js
```diff
@@ -0,0 +1,14 @@
+const usersControllers = require('./controllers/users')
```
des-des: I think singular for these names.

In the same way the model is called User, not Users,

Call this userController

src/routes.js
```diff
@@ -0,0 +1,14 @@
...
+
+const router = require('express').Router()
+
+// user routes
+router.route('/users')
+  .get(usersControllers.get)
+  .post(usersControllers.post)
```
des-des: Call this one `userControllers.create`.

If you feel strongly about that do not worry

src/controllers/users.js
```diff
@@ -0,0 +1,41 @@
...
+const usersHandlers = module.exports = {}
+
+usersHandlers.get = (req, res) => {
+  // sends back array of users, filtered by queries
+  // status codes: 200 (success)
+  User.find()
+    .then(users => {
+      res.send(users)
+    })
+    .catch(() => {
```
des-des: You are throwing away lots of information here
```js
.catch(error => {
  console.error(error)
  res.status(500).send({ message: 'database error: ' + error.message })
}
```

des-des (changes requested):

tests/controllers/users.test.js
mattlub: it's fine not to test the 500?

src/controllers/users.js
mattlub: what other stuff would you send? @des-des

src/routes.js
mattlub: For the record, I disagree but I think we're changing it anyway

src/controllers/users.js
des-des: The `error.message`

des-des (approved):

Issue #56, opened 2017-07-18
GET /users/:id tests and implementation


mattlub:
relates #48

PR #57, opened 2017-07-18
userController.getById /users/:id

mattlub:
implementation for controller for /users/:id GET endpoint

relates #56

### DIFF
--- a/src/controllers/users.js
+++ b/src/controllers/users.js
@@ -18,6 +18,7 @@ usersHandlers.getById = (req, res) => {
   // receives id in url
   // sends back one user
   // status codes: 200 (success), 404 (not found)
+  res.send('todo')
 }
 
 usersHandlers.post = (req, res) => {
--- a/src/controllers/users.js
+++ b/src/controllers/users.js
@@ -6,7 +6,7 @@ usersHandlers.get = (req, res) => {
   // sends back array of users, filtered by queries
   // status codes: 200 (success)
   User.find()
-    .then((users) => {
+    .then(users => {
       res.send(users)
     })
     .catch(err => {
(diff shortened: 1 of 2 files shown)


tests/controllers/users.test.js
```diff
@@ -29,8 +29,37 @@ tape('test /users GET returns list of users', t => {
...
+    .expect(404)
+    .expect('Content-Type', /json/)
+    .end((err, res) => {
+      if (err) t.fail(err)
+      t.equal(res.body.message, 'Cannot find user with id=10', 'response message should be "Cannot find user with id=10"')
+      dropCollectionAndEnd(User, t)
+    })
+})
+
+tape('test /users/:id GET with id of something in the database', (t) => {
```
m4v15, line 53: shall we change the req parameter to _id so ```/users/:_id``` as this is how we need it to look for mongoose
@mattlub

src/controllers/users.js
```diff
@@ -19,6 +19,14 @@ userControllers.getById = (req, res) => {
   // receives id in url
   // sends back one user
   // status codes: 200 (success), 404 (not found)
+  const id = req.params.id
+  User.findById(id)
+    .then(user => res.send(user))
+    .catch(() => {
+      // probably change this error soon to use the message from the db error
+      const errorObj = { message: `Cannot find user with id=${id}` }
+      return res.status(404).send(errorObj)
```
m4v15: why return?

tests/controllers/users.test.js
m4v15: dw

des-des (approved):

des-des (approved):

Issue #59, opened 2017-07-18
PUT /users/:id tests + implementation


mattlub:
relates #48

PR #61, opened 2017-07-18
User update controller

mattlub:
relates #59 #48

### DIFF
--- a/src/controllers/users.js
+++ b/src/controllers/users.js
@@ -34,8 +34,12 @@ userControllers.update = (req, res) => {
   // amends db record
   // status codes: 200 (success), 400 (bad request)
   const id = req.params.id
-  const errorObj = { message: 'Database error' }
-  res.status(400).send(errorObj)
+  User.findByIdAndUpdate(id, req.body, { new: true })
+    .then(updatedUser => res.send(updatedUser))
+    .catch(err => {
+      const errorObj = { message: `Database error: ${err.message}` }
+      res.status(400).send(errorObj)
+    })
 }
 
 userControllers.delete = (req, res) => {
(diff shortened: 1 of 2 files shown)


m4v15 (approved):

src/controllers/users.js
```diff
@@ -32,6 +32,13 @@ userControllers.update = (req, res) => {
   // receives updated json for user in body
   // amends db record
   // status codes: 200 (success), 400 (bad request)
+  const id = req.params.id
+  User.findByIdAndUpdate(id, req.body, { new: true })
+    .then(updatedUser => res.send(updatedUser))
+    .catch(err => {
+      const errorObj = { message: `Database error: ${err.message}` }
+      res.status(400).send(errorObj)
```
des-des, line 40: Should be 500?

des-des: Hmm okay, it seems findByIdAndUpdate returns an error if user does not exist, in this case should be 404, in other cases prop 500.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

tests/controllers/users.test.js
```diff
@@ -42,5 +42,34 @@ tape('test /users GET returns list of users', t => {
 // Tests for: POST /users
 
 // Tests for: PUT /users/:id
+tape('test PUT request to /users/:id with id of something not in the database', (t) => {
+  supertest(server)
+    .put('/users/invalidId')
+    .send(validUser1)
+    .expect(400)
```
des-des, line 49: 400 is bad request, 404 is not found

des-des (changes requested):

des-des (approved):

Issue #71, opened 2017-07-20
Create error handling middleware


des-des:
- [x] create middleware for error handling
   - [x] add basic middleware
   - [x] add if statements to deal with most common errors
      - [x] invalid id
      - [x] validation errors
      - [x] no id found in database
        - [x] add static methods to the schemas to error when id is not found (atm it just returns null)
      - [x] fails unique constraint
      - [x] don't hard code in the error codes or names, have a JSON somewhere.
 - [x] test the middleware
 - [x] refactor old code
   - [x] pass ```next``` to each route, call it on the ```catch```
   - [ ] loose some of the tests

des-des:
@m4v15 @mattlub is there anything outstanding on this?

m4v15:
No, all done. We could I guess lose some tests, but I (we) think they're good integration tests (in the controllers file)

PR #74, opened 2017-07-23
Implement error handling using custom middleware

m4v15:
NO LONGER A WIP

- tidies up the 4 controllers (they are all identical)
- changes some of the tests as the error messages have changed.
- adds package-lock.json
- add handling of mongo duplicate key error

N.B. tests could be cleaned up, but I think this is a separate job- e.g. abstracting out the tests as they are all the same.

relates #71


m4v15:
as long as you don't get the contributions for the package-lock.json, I'm happy

des-des (approved):

PR #76, opened 2017-07-25
add Db utils

mattlub:
add db utils.js file, containing custom static methods for Mongoose schema, which error if they don't find anything matching the given id.

add these methods to the models.

relates #71

currently 'failing' some checks because codecov is down I think, can be merged anyway.

### DIFF
--- a/src/db/utils.js
+++ b/src/db/utils.js
@@ -1,6 +1,8 @@
+const { messages: errMessages, names: errNames } = require('../constants/errors')
+
 const createCustomDbError = message => {
   const err = new Error(message)
-  err.name = 'CustomDbError'
+  err.name = errNames.CUSTOM
   return err
 }
 
@@ -9,7 +11,7 @@ const findByIdOrError = function (id) {
   return this.findById(id)
     .then(res => {
       if (res === null) {
-        return Promise.reject(createCustomDbError('No document matching that id'))
+        return Promise.reject(createCustomDbError(errMessages.GET_ID_NOT_FOUND))
       }
       return Promise.resolve(res)
     })
@@ -21,7 +23,7 @@ const findByIdAndUpdateOrError = function (id, data, options) {
   return this.findByIdAndUpdate(id, data, options)
     .then(res => {
       if (res === null) {
-        return Promise.reject(createCustomDbError('Cannot find document to update'))
+        return Promise.reject(createCustomDbError(errMessages.UPDATE_ID_NOT_FOUND))
       }
       return Promise.resolve(res)
     })
@@ -33,7 +35,7 @@ const findByIdAndRemoveOrError = function (id, data, options) {
   return this.findByIdAndRemove(id)
     .then(res => {
       if (res === null) {
-        return Promise.reject(createCustomDbError('Cannot find document to delete'))
+        return Promise.reject(createCustomDbError(errMessages.DELETE_ID_NOT_FOUND))
       }
       Promise.resolve(res)
     })
--- a/src/db/utils.js
+++ b/src/db/utils.js
@@ -8,7 +8,7 @@ const createCustomDbError = message => {
 const findByIdOrError = function (id) {
   return this.findById(id)
     .then(res => {
-      if (!res) {
+      if (res === null) {
         return Promise.reject(createCustomDbError('No document matching that id'))
       }
       return Promise.resolve(res)
@@ -20,7 +20,7 @@ const findByIdOrError = function (id) {
 const findByIdAndUpdateOrError = function (id, data, options) {
   return this.findByIdAndUpdate(id, data, options)
     .then(res => {
-      if (!res) {
+      if (res === null) {
         return Promise.reject(createCustomDbError('Cannot find document to update'))
       }
       return Promise.resolve(res)
@@ -32,7 +32,7 @@ const findByIdAndUpdateOrError = function (id, data, options) {
 const findByIdAndRemoveOrError = function (id, data, options) {
   return this.findByIdAndRemove(id)
     .then(res => {
-      if (!res) {
+      if (res === null) {
         return Promise.reject(createCustomDbError('Cannot find document to delete'))
       }
       Promise.resolve(res)
--- a/src/db/utils.js
+++ b/src/db/utils.js
@@ -6,44 +6,38 @@ const createCustomDbError = message => {
 
 // static method for Mongoose schema
 const findByIdOrError = function (id) {
-  return new Promise((resolve, reject) => {
-    this.findById(id)
-      .then(user => {
-        if (!user) {
-          return reject(createCustomDbError('No document matching that id'))
-        }
-        resolve(user)
-      })
-      .catch(reject)
-  })
+  return this.findById(id)
+    .then(res => {
+      if (!res) {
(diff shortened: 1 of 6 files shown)


m4v15 (approved):

m4v15:
Again, methods are registered not used, so coverage has gone done

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
+const createCustomDbError = message => {
+  const err = new Error(message)
+  err.name = 'CustomDbError'
```
des-des: Yeah, import this in

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+  return err
+}
+
+// static method for Mongoose schema
+const findByIdOrError = function (id) {
+  return new Promise((resolve, reject) => {
```
des-des: findById seems to return a promise... The promise wrapper is unessisary

```js
findByIdOrError = id => this.findById(id)
  .then(user => {
    if (!user) {
      return Promise.reject(createCustomDbError('No document matching that id'))
    }
    return Promise.resolve(user)
})
```

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+        }
+        resolve(user)
+      })
+      .catch(reject)
+  })
+}
+
+// static method for Mongoose schema
+const findByIdAndUpdateOrError = function (id, data, options) {
+  return new Promise((resolve, reject) => {
```
des-des: same as above

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+  return err
+}
+
+// static method for Mongoose schema
+const findByIdOrError = function (id) {
+  return new Promise((resolve, reject) => {
+    this.findById(id)
+      .then(user => {
+        if (!user) {
+          return reject(createCustomDbError('No document matching that id'))
```
des-des: string from constants

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+  err.name = 'CustomDbError'
+  return err
+}
+
+// static method for Mongoose schema
+const findByIdOrError = function (id) {
+  return new Promise((resolve, reject) => {
+    this.findById(id)
+      .then(user => {
+        if (!user) {
```
des-des: Can you check explicitly. Ie `if (user === undefined)` or something

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+  })
+}
+
+// static method for Mongoose schema
+const findByIdAndUpdateOrError = function (id, data, options) {
+  return new Promise((resolve, reject) => {
+    this.findByIdAndUpdate(id, data, options)
+      .then(user => {
+        if (!user) {
+          return reject(createCustomDbError('Cannot find document to update'))
```
des-des: same as above

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+      .catch(reject)
+  })
+}
+
+// static method for Mongoose schema
+const findByIdAndUpdateOrError = function (id, data, options) {
+  return new Promise((resolve, reject) => {
+    this.findByIdAndUpdate(id, data, options)
+      .then(user => {
+        if (!user) {
```
des-des: same as above

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+        }
+        resolve(user)
+      })
+      .catch(reject)
+  })
+}
+
+// static method for Mongoose schema
+const findByIdAndRemoveOrError = function (id, data, options) {
+  return new Promise((resolve, reject) => {
```
des-des: same as above

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+    this.findByIdAndRemove(id)
+      .then(user => {
+        if (!user) {
```
des-des: same as above

src/db/utils.js
```diff
@@ -0,0 +1,53 @@
...
+          return reject(createCustomDbError('Cannot find document to delete'))
```
des-des: same as above

src/db/utils.js
des-des: Also, can you not call this user?

src/models/Event.js
```diff
@@ -20,4 +21,9 @@ const eventSchema = mongoose.Schema(
   }
 )
 
+// methods which throw errors when there's nothing matching the given id
+eventSchema.statics.findByIdOrError = findByIdOrError
```
des-des: Okay can you build a helper that does this, ie `addStaticMethods`

src/models/Place.js
```diff
@@ -22,4 +23,9 @@ const placeSchema = mongoose.Schema(
   }
 )
 
+// methods which throw errors when there's nothing matching the given id
+placeSchema.statics.findByIdOrError = findByIdOrError
```
des-des: use helper.

That way, when you add a new method, you do not need to do it everywhere..

des-des (changes requested):

mattlub:
cheers @des-des , updated

src/db/utils.js
```diff
@@ -6,44 +6,38 @@ const createCustomDbError = message => {
...
-      .catch(reject)
-  })
+  return this.findById(id)
+    .then(res => {
+      if (!res) {
+        return Promise.reject(createCustomDbError('No document matching that id'))
+      }
+      return Promise.resolve(res)
+    })
+    .catch(err => Promise.reject(err))
```
des-des: This does not do anything, you can omit it

src/db/utils.js
```diff
@@ -6,44 +6,38 @@ const createCustomDbError = message => {
...
-      .catch(reject)
-  })
+  return this.findByIdAndUpdate(id, data, options)
+    .then(res => {
+      if (!res) {
+        return Promise.reject(createCustomDbError('Cannot find document to update'))
+      }
+      return Promise.resolve(res)
+    })
+    .catch(err => Promise.reject(err))
```
des-des: same

src/db/utils.js
```diff
@@ -6,44 +6,38 @@ const createCustomDbError = message => {
...
-      .catch(reject)
-  })
+  return this.findByIdAndRemove(id)
+    .then(res => {
+      if (!res) {
+        return Promise.reject(createCustomDbError('Cannot find document to delete'))
+      }
+      Promise.resolve(res)
+    })
+    .catch(err => Promise.reject(err))
```
des-des: same

src/db/utils.js
des-des: like: I am gonna catch a rejected promise just so I can reject it again

Just seem mean to the error

des-des (changes requested):

des-des:
@mattlub @m4v15 

expand the points above

image

m4v15 (approved):

des-des:
@m4v15 @mattlub, ready to go, with one comment.

You need to expand my points above to see it

des-des (approved):

PR #80, opened 2017-07-26
Unit test error handlers

m4v15:
relates #71


tests/dbErrorHandlers.test.js
```diff
@@ -0,0 +1,49 @@
+const dbErrorHandlers = require('../src/dbErrorHandlers.js')
+const { messages: errMessages, names: errNames, codes: errCodes } = require('../src/constants/errors.json')
+const { buildNextSpyTest, buildBoomSpyTest } = require('./helpers/index.js')
+
+const tape = require('tape')
+
+tape('dbErrorHandlers.custom with non custom error', t => {
+  buildNextSpyTest(t, dbErrorHandlers.custom, 'test error message', 'not a custom error')
```
des-des: Okay, this is quite hard to read, or work out what the hell is going on, yes, it is good to keep tests dry, but also they should be very easy to read.

des-des (changes requested):

tests/dbErrorHandlers.test.js
des-des: Okay,

I think:
```js
callErrorMiddleware = errorMiddleware => (error) => {
  const resSpy = 
  const nextSpy = 
  errorMiddleware(error, undefined, resSpy, nextSpy)
}
```

Then in your tests
```js
test('testing something', t => {
  const callMiddleware = callErrorMiddleware(middlewareToTest)

  const errorToTest = some error
  const { resSpy, nextSpy } = callMiddleware(errorToTest)

  // now, in the test function, we can make assertions in plain English
})
```

Okay that is very rough, but the idea is that you do not really need to look in the helpers to understand what assertions are being made. The helper should have names that makes it obvious what they are doing, and the assertions want to live in the test.

If you want to make helpers that make assertions, the most you want to be doing is having functions that are called in tests like
```js
assertSomethingRepetive(t, actual, expected, msg)
```

des-des (changes requested):

m4v15:
@des-des I've changed it now to be more like what you said. It's fairly repetitive, but I think in a good way to make the tests more readable.

Not sure on my function name, but it's sort of growing on me now...

tests/helpers/index.js
```diff
@@ -1,7 +1,28 @@
...
+    }
+  }
+  return resSpy
+}
+
+helpers.spyGeneratorErrorMiddlewareCaller = errorMiddleware => (error) => {
+  const resSpy = buildResponseObj()
+  const nextSpy = sinon.spy()
+  errorMiddleware(error, undefined, resSpy, nextSpy)
+  return { resSpy, nextSpy }
```
des-des, line 27: :+1: :

tests/helpers/index.js
```diff
@@ -1,7 +1,28 @@
...
+    boom: {
+      badRequest: sinon.spy(),
+      notFound: sinon.spy(),
+      badImplementation: sinon.spy()
+    }
+  }
+  return resSpy
+}
+
+helpers.spyGeneratorErrorMiddlewareCaller = errorMiddleware => (error) => {
```
des-des, line 23: I like the name

des-des (changes requested):
Okay I have one change to request, it kind long though.

Make a folder called middlewar, then put each of the middlewares in its own file.

Finally mirror this in you test files.

Sorry not to have asked earlier.. Hope that okay!

look good though good work

des-des (approved):

PR #86, opened 2017-07-27
add docs (work in progress)

jwld:
relates #73

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -11,7 +11,7 @@ An API providing data for the four apps on the Nazareth Open Tourism Platform, b
 
 ## Events
 
-#### Get all events
+### Get all events
 `GET /events`
 
 **Response**
@@ -40,7 +40,7 @@ Status: 200 OK
 ]
 ```
 
-#### Get event by id
+### Get event by id
 `GET /events/:id`
 
 **Response**
@@ -67,7 +67,7 @@ Status: 200 OK
 }
 ```
 
-#### Create event
+### Create event
 `POST /events`
 
 **Input**
@@ -107,7 +107,7 @@ Status: 201 Created
 }
 ```
 
-#### Update event
+### Update event
 `PUT /events/:id`
 
 **Input**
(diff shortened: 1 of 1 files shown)


docs.md
```diff
@@ -0,0 +1,157 @@
+# Open Tourism Platform API
+An API providing data for the four apps on the Nazareth Open Tourism Platform, built by graduates of [Founders and Coders](https://foundersandcoders.com/about/).
+
+## Contents
+**Events**
+- [Get all events](#Get all events)
```
m4v15: these links don't work

m4v15: need no capitals and hyphen instead of spaces in the brackets I think

jwld: thanks, fixed 👍

docs.md
```diff
@@ -0,0 +1,157 @@
...
+### Create event
+`POST /events`
+
+**Input**
+
+Name | Type | Description
+---|---|---
+name | string | **Required**. The name of the event.
+category | array of strings | **Required**. The event category / categories.
+accessibilityOptions | array of strings | Event accessibility options.
```
des-des, line 79: I think these may be restricted.

See here: https://github.com/foundersandcoders/open-tourism-platform/blob/master/src/models/constants.json

These will prob change, for now you could just drop in a link

docs.md
```diff
@@ -0,0 +1,157 @@
...
+  "category": [
+    "music"
+  ]
+}
+```
+
+### Create event
+`POST /events`
+
+**Input**
```
des-des, line 73: Github called them parameters.

https://developer.github.com/v3/repos/#parameters

Not sure what is better

jwld: Create and update both refer to them as 'input' however: 

https://developer.github.com/v3/repos/#create

des-des: :+1:

des-des (approved):

m4v15:
@des-des, it was work in progress, it wasn't right

m4v15:
The objects that are returned now have ```en``` and ```ar``` keys for exampe

des-des:
derp

des-des:
sorry

Issue #27, opened 2017-07-05
Authentication


mattlub:
UPDATE - Making this the Epic for Auth 14/08/17
----
We have spent sprint 2 setting up the auth, but as we didn't know exactly what we were doing we haven't raised issues particularly well or documented our progress. I am now going to attempt to break up what we have done so far and what more we need to do. This epic issue can then be broken down.

## How Authorisation and Authentication will work on OTP

1. Anyone can register to be a user on OTP, will get a default `role` of `BASIC`
2. Routes will be secured by checking a users `role` and seeing if they are allowed on that route.
3. Further fine level permissioning (i.e. what fields different users can edit) will be also be done.
4. Apps will be able to make requests on behalf of a user once they have been authorised by that user. The app will have the same level of permissions as the user who authorised them.

### User Authentication / sessions

1. User logs in, has details added to cookie
2. Secure routes check cookie and scope of the route, let through or don't let through

More detail in #100 and have begun to be addressed in #95 

### OAuth
- Users [oath2-server](https://github.com/oauthjs/node-oauth2-server)
1. Superuser registers apps on the platform, giving them a client_id and a client_secret and a redirecitURI
2. App directs user to `oath/authorize` for user to authorise the app on their behalf (with the app's client id and redirect URI)
3. If the user authorises the app, it will make a `POST /oath/authorize` request. oath2-server will then use a (custom) authenticate handler to return the correct user, and then will redirect back to the redirect URI with an authorisation code(temp)
4. This redirect URI (within the app) will need to make `POST /oath/token` request, with the client id, client secret and the temp auth code, and will be sent back a permaneant token.
5. The app can then store this token somewhere and use it to make requests on the users behalf.
6. The secure routes will now check if there is a JWT in the cookie, on an auth token in a header, allowing them through as though they are the user (or have the users role) either way.

Initial steps addressed in #92.

### Permissioning

Yet to be started, but this is actually the point of all the auth. Can refer to the spec and lots of previous discussions about who can do what, will mostly be just checking the `role` or the `userid` (to see if they're the owner).

## To do:

- [x] Finish sessions #100
- [x] Write the custom Authentication handler for oauth2-server
- [x] add middleware that looks for oath token in header
- [x] Add views for UI #104 
  - [x] Set up handlebars and tachyons
  - [x] Add and serve views for login, register user, authorise app routes (`GET` requests)
- [x] Add the authentication middleware into the appropriate routes
[...]

PR #92, opened 2017-08-02
initial OAuth set up

m4v15:
#### ready for review.
@mattlub comments:
- added models
- adds one integration test
- there is a dummy authenticate function

----

**OLD comments:**

DO NOT MERGE
Not going to be ready for a while just trying out a few things

relates #93 #27 

UPDATE from @mattlub:
I've done a fair bit of work on this,
**see my comments below**


mattlub:
I've picked this up. Have refactored a fair bit.

Instructions on how to see what I've done:
**TODO** (message me in the meantime)

I've been working with the following 3 links open:
http://oauth2-server.readthedocs.io/en/latest/api/oauth2-server.html#oauth2server-token
https://tech.zilverline.com/2017/03/17/nodejs-oauth2-provider
https://github.com/slavab89/oauth2-server-example-mongodb

### **definitions:**
**user** - user of the platform, a person, or 'resource owner'
**client**- an app, owned by a user

### reminder of the flow
- an authenticated user grants access to a client, saying that client can access the user's platform data
- the platform issues an `authorization code` to the client, specific to the user
- the client requests an `access token` from the platform for the specific user

## routes:
#### `GET /oauth/authorize`
- **required query params: `client_id`, `redirect_uri`**
- should render a login page for an individual user to log in with the platform
- if logged in, or after login, should render a page asking the user to grant access to its data to the specific client
- upon granting access, should sent a request to `POST /oauth/authorize`

#### `POST /oauth/authorize` 
- it requires an `Authentication: bearer {token}` header, to assure the grant for access has come from the actual user
- it has the following url params: `client_id=...`, `redirect_uri=...`, `response_type=code`, `state=<randomstring>`
- it should redirect to the `redirect_uri` including a url param `code=<authorization_code>`

#### `POST /oauth/token`
- requires a `Content-Type: application/x-www-form-urlencoded` header
- sends `client_id`, `client_secret`, `code=<authorization_code>`, `grant_type=authorization_code`
- should get back token, this is not working yet.


**TODO:**
- change UserTests to use User
- make working tests
- complete token request

mattlub:
required views/endpoints:
- `login` for users
- `my apps` page for users
- `add app` page/route for users
- `authorize` view for each specific app presented to users

more specifically, required endpoints:
- `POST /oauth/clients`: requires user authentication, create new client, provide redirect_uri, generate secret
- TODO

des-des:
Okay I have done some reading, and want to try and drop in a top level description of how this should all fit together.

In terms of authentication of requests made to the OTP API.
A request can be authenticated in one of two ways.
  1. An app acting on behalf of a user will provide an authentication token. Here is an example of middleware that does this: https://github.com/oauthjs/express-oauth-server/blob/master/index.js#L42
  2. A user who is interacting directly with OTP, will provide a JWT, issued by us, this will be verified using the jwt secret and we will use the claims to extract the user data from the database.

Now, we number 2 basically set up. Creating the authentication token for part 1 is the (very) difficult part.

The module we are using will provide 3 methods, once we have configured it to properly interface with the database.
  1. authenticate: this will be used in our implementation of step 1 above
      1. Extract the authentication token from the authentication header
      2. retrieve the user who granted the token
      3. Check that the token has not expired
  2. Authorise: Here we accept a client_id from the body of a request and a redirect_url, authorise will generate a short lived authorisation code, which is embedded in the redirect. Here authorise is handling the response, so the user gets redirected without any further action on our part. @mattlub **A user must be authenticated at this point**.
  3. token. This handles a request that contains a client_secret, a client_id and an authorisation code. As far as I can see, this will handle the server response with a access token in the body.

So there is a kinda hidden step in here, that is not immediately obvious from looking at those three functions, but might be the missing piece in terms of understanding. Here is I think a key point https://github.com/oauthjs/node-oauth2-server/issues/264#issuecomment-193209294!

Basically, there are two steps in the granting of the authorisation code
  1. Login
  2. grant access

The second step here utilises the authorise function, with a custom authorisationFunction. There is another way to do this, but I do not understand it and it seems like passing a custom authorisationFunction makes more sense.

Here is an example:
https://github.com/ubilogix/koa2-oauth-server/blob/master/examples/index.js#L273
with login redirecting back to grant page https://github.com/ubilogix/koa2-oauth-server/blob/master/examples/index.js#L226

## Next steps
I see the best way forward being
  1. Set up a fake user in the database, create a custom authorisationFunction that assumes that user is logged in
  2. Finalise this pull with the above hack, without any of the views, just integration tests
  3. Update the other pull to use cookies instead of headers to store the session
  4. Once both pulls are together, update the authorisationFunction to make use of the session functionallity we have (see example above)
  5. Finally, we need to build the views, then move onto implementing the authentication layer in the app API.

mattlub:
@des-des 
>  Here is I think a key point https://github.com/oauthjs/node-oauth2-server/issues/264#issuecomment-193209294!

thanks, this was useful, flagging up similar things which were causing me problems, I think the module is just designed right now in a non-standard way.

I think we are on the same page now, more or less.

---

**Next steps in my opinion:**
- you created a custom user authenticate handler in your other branch, this should be updated in some way to deal with access_tokens as well.
- we need to add this custom authenticate handler as the auth handler for the `/authorize` route, as described [here](https://tech.zilverline.com/2017/03/17/nodejs-oauth2-provider)

des-des:
@mattlub yup you are correct! But I see those things coming between point 3 and 4 above.

Ie now we know what we are doing, we need to find a way of moving stuff into master without merging everything together into one giant pull..

mattlub:
@des-des yeh sorry, I hadn't seen your 'next steps' section when I wrote my comment. I'm good to start on your steps 1 and 2 tomorrow.

### My comments/issues/things I'd like your opinion on:
1. **Do we store normal auth tokens in different table from OAuth access tokens?** They are created in completely different flows, but could be the same model.
1. **how does the way we deal with authenticating requests from clients and users differ?** In the simplest cases where apps have full permissions for users I guess there doesn't have to be much of a difference, but in more complicated cases I haven't really thought about it. I guess just middleware, which results in a scope object.

des-des:
@mattlub

Clients will not authenticate.

A user can do 2 things:
  1. Create a app / client
  2. Grant a client permission to act on behalf of them (for mvp I think we will have only one scope, apps will not get granted super user scope, but will be able to do everything else the user can do)

Ie a super user might both these things, both allow another app to act on their behalf, but also create their own app (think github).

Not sure I am quite following you, but there will be 2 different ways to authenticate.
  1. We have header `authorisation=bearer {oauthToken}`
  2. I think we will move jwt we currently have into cookie,ie jwt in cookie (right now it is in header)

I do not think we need to store jwts right now, we can give them a 1h expiry time, and not store them on the server.

Lets chat about anything that does make sense tomorrow morning (sorry maybe kinda rambling)

src/authDevRouter.js
```diff
@@ -0,0 +1,73 @@
+const router = require('express').Router()
```
m4v15: what's happening with this file? Do we want it kept/should we put it somewhere else

src/authDevRouter.js
```diff
@@ -0,0 +1,73 @@
...
+
+// DEVELOPMENT/TESTING ROUTES
+
+// empty the db
+router.get('/deleteAll', (req, res) => {
```
des-des: hmm, I do not like like this going into master, could you just make this a helper function that lives inside the test file somewhere

src/authDevRouter.js
```diff
@@ -0,0 +1,73 @@
...
+    .then(users => res.send(users))
+    .catch(err => res.send(err))
+})
+router.get('/codes', (req, res) => {
+  AuthorizationCode.find()
+    .then(codes => res.send(codes))
+    .catch(err => res.send(err))
+})
+
+module.exports = router
```
des-des: This file confuses me, why not just make helper methods that call the models directly and live inside helpers.

I do not like all these endpoints sitting in production code they seem like a bit of a liability.

Thoughts?

src/authModel.js
```diff
@@ -0,0 +1,89 @@
+const Token = require('./models/auth/Token')
+const Client = require('./models/auth/Client')
+const AuthorizationCode = require('./models/auth/AuthorizationCode')
+
+module.exports = {
+  getAccessToken: accessToken => {
+    console.log('finding token, accessToken = ' + accessToken)
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+    return Token.findOne({ accessToken })
+      .populate('user')
+  },
+
+  getAuthorizationCode: authCode => {
+    console.log('finding auth code, code = ' + authCode)
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+    return AuthorizationCode.findOne({ authorizationCode: authCode })
+      .populate('user client')
+  },
+
+  getClient: (clientId, clientSecret) => {
+    console.log('finding client, id = ' + clientId + ', secret = ' + clientSecret)
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+    const params = { _id: clientId }
+    if (clientSecret) params.secret = clientSecret
+    return Client.findOne(params)
+  },
+
+  getRefreshToken: refreshToken =>
+    Token.find({ refreshToken })
+      .populate('user'),
+      // .exec(),
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+
+  revokeAuthorizationCode: authCode => {
+    // should return true if successful
+    console.log('revoking auth code: ')
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+    return AuthorizationCode.findOneAndRemove({ authorizationCode: authCode.authorizationCode })
+      .then(doc => !!doc)
```
des-des: or `() => true`?

What is the significance of whether you get a doc, is there something else you might get instead?

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+  },
+
+  revokeToken: token => {
+    Token.findOneAndRemove({ refreshToken: token.refreshToken })
+      .then(doc => !!doc)
```
des-des: again, what are the possible returns?

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+  },
+
+  saveAuthorizationCode: (code, client, user) => {
+    console.log('saving authorization code.')
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+      .then(authCode => ({
+        authorizationCode: authCode.authorizationCode,
+        expiresAt: authCode.expiresAt,
+        redirectUri: authCode.redirectUri,
+        scope: authCode.scope,
+        client: {id: authCode.client},
+        user: {id: authCode.user}
+      })
+      )
+      .catch(err => err)
```
des-des: :fire:

src/authModel.js
```diff
@@ -0,0 +1,89 @@
...
+        accessToken: savedToken.accessToken,
+        accessTokenExpiresAt: savedToken.accessTokenExpiresAt,
+        refreshToken: savedToken.refreshToken,
+        refreshTokenExpiresAt: savedToken.refreshTokenExpiresAt,
+        scope: savedToken.scope,
+        client: {id: savedToken.client},
+        user: {id: savedToken.user}
+      })
+      )
+      .catch(err => err)
```
des-des: :fire:

src/authRouter.js
```diff
@@ -0,0 +1,59 @@
...
+const Client = require('./models/auth/Client')
+
+// create new OAuthServer
+const oauth = new OAuthServer({ model: authModel })
+const router = require('express').Router()
+
+// AUTHORIZE ROUTES
+
+// should display the authorization grant page for a specific app/client
+router.get('/authorize', (req, res, next) => {
```
des-des: refac?

Move into controllers / router?

src/authRouter.js
```diff
@@ -0,0 +1,59 @@
...
+  if (!req.query || !req.query.client_id) {
+    return res.boom.badRequest('no client_id provided')
+  }
+  Client.findOne({ id: req.query.client_id })
+    .populate('user')
+    .then(client => {
+      if (client === null) {
+        return res.boom.badRequest('client_id is incorrect')
+      }
+      console.log(`app created by ${client.user.username} requests your authorisation.`)
```
des-des: :fire:

src/authRouter.js
```diff
@@ -0,0 +1,59 @@
...
+      // TODO: this should display correct authorization grant prompt page
+      res.sendFile(path.join(__dirname, 'public', 'authorize.html'))
+    })
+    .catch(next)
+})
+
+// should redirect supplying an authorization code
+router.post('/authorize', oauth.authorize({
+  authenticateHandler: {
+    // dummy function for now, just finds a user
```
des-des: Can we add a comment
`// DANGER: Currently insecure`

src/middleware/mongooseErrorHandler.js
```diff
@@ -22,9 +22,9 @@ module.exports = (err, req, res, next) => {
       res.boom.badRequest(errMessages.INVALID_ID)
       break
 
-    // unhandled mongoose error
+    // unhandled
     default:
-      res.boom.badImplementation(errMessages.UNHANDLED_MONGOOSE)
+      next(err)
```
des-des, line 27: :+1:

src/middleware/finalErrorHandler.js
```diff
@@ -0,0 +1,6 @@
+const { messages: errMessages } = require('../constants/errors')
+
+module.exports = (err, req, res) => {
+  console.error(err)
+  res.boom.badImplementation(errMessages.UNHANDLED_CUSTOM)
```
des-des: I think Boom eats this error message. So you can leave the argument blank!

src/models/auth/Client.js
```diff
@@ -0,0 +1,13 @@
+const mongoose = require('mongoose')
+
+const clientSchema = mongoose.Schema(
+  {
+    id: String, // client_id
```
des-des: Is the comment necessary?

src/public/authorize.js
```diff
@@ -0,0 +1,37 @@
+document.addEventListener('DOMContentLoaded', function (event) {
+  var form = document.querySelector('#authorize-form')
+
+  form.addEventListener('submit', function (e) {
+    e.preventDefault()
+
+    console.log('custom form submit')
+
+    var headers = new Headers({
```
des-des: I think we are gonna have a cookie for the jwt.. What is this for?

src/public/authorize.js
```diff
@@ -0,0 +1,37 @@
...
+      'Authorization': 'Bearer token'
+    })
+
+    var init = {
+      method: 'POST',
+      headers: headers,
+      mode: 'cors',
```
des-des: not cors?

src/server.js
```diff
@@ -1,22 +1,27 @@
 const express = require('express')
 const router = require('./routes')
+const authRouter = require('./authRouter')
```
des-des: refac router

src/server.js
```diff
@@ -1,22 +1,27 @@
...
 const bodyParser = require('body-parser')
 const boom = require('express-boom')
+const path = require('path')
 
 const customErrorHandler = require('./middleware/customErrorHandler')
 const mongoErrorHandler = require('./middleware/mongoErrorHandler')
 const mongooseErrorHandler = require('./middleware/mongooseErrorHandler')
-
+const finalErrorHandler = require('./middleware/finalErrorHandler')
```
des-des, line 11: good

src/server.js
```diff
@@ -1,22 +1,27 @@
...
 const server = express()
 
+server.use(express.static(path.join(__dirname, 'public')))
 server.use(bodyParser.urlencoded({ extended: true }))
 server.use(bodyParser.json({ extended: true }))
 server.use(boom())
 
-server.use(router)
+server.use(router) // API endpoints
+server.use('/oauth', authRouter)
```
des-des: Okay, happy to actually have 2 routers, apiRouter and authRouter.

While we are at it, you could move add `/api/v1` to the api router, but db if that is too much faff

tests/OAuth/index.test.js
```diff
@@ -0,0 +1,16 @@
+require('../../config.js')
+const tape = require('tape')
+const dbConnection = require('../../src/db/connect.js')
+
+dbConnection.once('open', () => {
```
des-des: okay can we move these tests into the normal test setup now?

tests/OAuth/integration.test.js
```diff
@@ -0,0 +1,104 @@
...
+  // do the test
+  .then(() => {
+    const randomState = 'randomString'
+    supertest(server)
+      .post('/oauth/authorize')
+      // .set('Authorization', 'bearer token')
+      .query({
+        client_id: client._id,
+        redirect_uri: client.redirectUris[0],
+        state: randomState,
```
des-des: what is state?

tests/OAuth/integration.test.js
```diff
@@ -0,0 +1,104 @@
...
+  Promise.resolve()
+  // add valid user, client
+  .then(() => User.create(user))
+  .then(() => Client.create(client))
+  // do the test
+  .then(() => {
+    const randomState = 'randomString'
+    supertest(server)
+      .post('/oauth/authorize')
+      // .set('Authorization', 'bearer token')
```
des-des: :fire: ?

tests/fixtures/users.json
```diff
@@ -1,7 +1,16 @@
 {
-   "validUser1": {"username":"mattlub","password":"pword", "en":{"name":"matt"},"email":"xxx","role":"SUPER","isPublic":false},
-   "validUser2": {"username":"m4v15","password":"pword","ar":{"name":"mavis"},"email":"xxx","role":"SUPER","isPublic":false},
-   "invalidUser1": {"username":"des-des","password":"pword", "email":"xxx","role":"SUPER","isPublic":false}
-
+	"validUser1": {"username":"mattlub","password":"pword", "en":{"name":"matt"},"email":"xxx","role":"SUPER","isPublic":false},
```
des-des, line 2: can we beautify this/

tests/OAuth/integration.test.js
```diff
@@ -0,0 +1,104 @@
...
+// Tests for: POST /oauth/authorize
+tape('POST /oauth/authorize should successfully redirect and create authorization code', t => {
+  Promise.resolve()
+  // add valid user, client
+  .then(() => User.create(user))
+  .then(() => Client.create(client))
+  // do the test
+  .then(() => {
+    const randomState = 'randomString'
+    supertest(server)
```
des-des: we can flatten this one level, ie `return supertest(...`

You can put random state one level out also

des-des (changes requested):
@mattlub Massive congrats on getting this in!

Sorry quite a few comments, hit me up if you disagree with anything!

src/authModel.js
m4v15: http://oauth2-server.readthedocs.io/en/latest/model/spec.html#revokeauthorizationcode-code-callback

true if code found and revoked, false if code not found

m4v15: Sorry I thought you were asking about what are the returns from the function being defined here not the findOneAndRemove function, which is a version of [findAndModify](https://docs.mongodb.com/manual/reference/command/findAndModify/#findandmodify-value-field)

If the document isn't found it will return value as null - so `() => true` would be wrong?

src/authModel.js
m4v15: ☝️

src/middleware/finalErrorHandler.js
```diff
@@ -0,0 +1,6 @@
+const { messages: errMessages } = require('../constants/errors')
+
+module.exports = (err, req, res) => {
```
m4v15: from the [express website (see the error handler section)](https://expressjs.com/en/guide/using-middleware.html):
>Error-handling middleware always takes four arguments. You must provide four arguments to identify it as an error-handling middleware function. Even if you don’t need to use the next object, you must specify it to maintain the signature. Otherwise, the next object will be interpreted as regular middleware and will fail to handle errors.

Other wise it assumes it's normal middleware so `(err, req, res)` becomes `(req, res, next)` and you get a `res.boom` doesn't work (which is how I discovered all this)

src/models/auth/Client.js
m4v15: Again, I put this in - I think it was mostly for clarity in terms of using the terms client id and client secret all the time this was just making it clear, but yeah not really necessary

src/authModel.js
m4v15: Just a heads up, I can't work out what happens when you do `.populate()` without `.exec()` because this is the problem you were having before which was only fixed by adding `.exec()`, @mattlub? I get this still passes the tests but I don't get why - have a feeling it's to do with only the client.id being the important part so maybe that can happen without `exec()` - everywhere I've looked always uses `.exec()` after `.populate()`

m4v15: Ok so I just manually checked this and it works it the original case without `exec()` so not really sure what the problem was before. I *think* it's to do with chaining queries but it doesn't matter if you have populate at the end.

Either way, ignore me

src/authModel.js
mattlub: yeh, mongoose returns the doc if it deleted one, else null

src/server.js
mattlub: I did this, but then all the tests failed 😃

src/middleware/finalErrorHandler.js
mattlub: cool thanks. I was wondering how it knows

tests/fixtures/users.json
mattlub: you mean with a plugin? Feel free to do it, I don't have a plugin.

tests/OAuth/integration.test.js
mattlub: see part 1 [here](https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-authorization-options-for-oauth-apps/)

"An unguessable random string. It is used to protect against cross-site request forgery attacks."

It is required by the module

tests/OAuth/integration.test.js
mattlub: what do you mean by this? put everything into one test?

src/authRouter.js
mattlub: how would you do it?

src/models/auth/Client.js
mattlub: there is slight issue/confusion on my end here as mongoose gives an _id value.

m4v15: yeah, tbh I'm not sure if it's needed, I feel like it might be due to the oauth module needing an id field specifically so maybe _id won't work, but I couldn't really test this at the time

tests/OAuth/integration.test.js
des-des: Derp ignore this

des-des:
@mattlub is this ready for me to look at again?

m4v15:
I think yes (from his gitter message) the only thing I questioned was adding the API mongoose models to their own folder and we weren't sure where the best place to put `Authmodel.js` cos it's not a mongoose model, but in the src folder seems a bit weird.

mattlub:
@des-des yes has been for 2 days!

des-des (approved):

Issue #89, opened 2017-07-29
populating fields, e.g. location in events document


mattlub:
I think we need to populate some of the fields, e.g. in events we should send back data about the location and not just the location id.

quite important imo, thinking about nazareth events.

m4v15:
I mean, it would just be a case of having to make 2 requests instead of one at them moment tho, right?

m4v15:
Also, what do you mean?

If I have an events database with a load of placeId's then run this:

`    Event.find().populate('placeId').exec().then(console.log).catch(console.log)`

It will return all of the events but with the placeId populated by the places table

If I just do this `    Event.find().then(console.log).catch(console.log)`

I'll just get it with placeId as the id

@mattlub

mattlub:
@m4v15 in the case of loading a list of events, *each* event would require another request

mattlub:
> It will return all of the events but with the placeId populated by the places table

@m4v15 yes, this is what I think should happen. But i couldn't get it to work using `event.find().populate`

m4v15:
you need the `.exec()` on the end

m4v15:
Also I'm doing the `.find` on the model, not on the individual document

mattlub:
Does it work?

m4v15:
Yeah, it does for me

m4v15:
I mean, obviously you need to make sure you have something in the database with that PlaceId - if you don't it'll just be replaced with null

m4v15:
```
  { _id: 597f0faa1d67c20a641f9d30,
    updatedAt: 2017-07-31T11:08:26.724Z,
    createdAt: 2017-07-31T11:08:26.724Z,
    en: { name: 'Party at the Basilica' },
    placeId: 
     { _id: 597f0faa1d67c20a641f9d2f,
       updatedAt: 2017-07-31T11:08:26.706Z,
       createdAt: 2017-07-31T11:08:26.706Z,
       en: [Object],
       __v: 0,
       accessibilityOptions: [],
       categories: [] },
    __v: 0,
    accessibilityOptions: [],
    categories: [ 'music' ] },
  { _id: 597f44acbe41160ef01863c7,
    updatedAt: 2017-07-31T14:54:36.512Z,
    createdAt: 2017-07-31T14:54:36.512Z,
    en: { name: 'Raving fun times everywhere' },
    placeId: null,
    __v: 0,
    accessibilityOptions: [],
    categories: [ 'music' ] } ]
```

[...]

Issue #81, opened 2017-07-26
sorting of returned items from GET requests


mattlub:
needs to be addressed, ideally should have sort related query params.

m4v15:
This got sorted right, @mattlub?

mattlub:
was done for events but not other routes. Also sort-related query params (e.g. order=desc) were not done.

PR #94, opened 2017-08-03
Change GET events endpoints

mattlub:
**UPDATE 14/8/17: READY FOR REVIEW**

This has turned into a bigger PR than intended, but not too much deep stuff going on.

Includes 
- [x] populate placeId field with place information for getAll and getById
- [x] sort events by date in getAll
- [x] convert controllers to use new `rejectIfNull` function, remove old custom schema methods
- [x] update docs

doesn't include (can come in another PR):
- [ ] change `placeId` field to `place` (?) #96 
- [ ] date-related query params for events

-----

after discussing, I think these need to be changed, as Mohamed is working on the nazareth-events and it is already causing problems.

relates #81 #89 

notes (outdated):
- brings up need for dropCollection**s**AndEnd
- cannot implement population easily for getById because of the helper functions in db Utils
- docs will need to be updated

### DIFF
--- a/src/controllers/event.js
+++ b/src/controllers/event.js
@@ -1,5 +1,5 @@
 const Event = require('../models/Event')
-const { errorIfNotFound } = require('../db/utils')
+const { rejectIfEmpty } = require('../db/utils')
 const { messages: errMessages } = require('../constants/errors')
 
 const eventController = module.exports = {}
@@ -17,9 +17,9 @@ eventController.getById = (req, res, next) => {
   // receives id in url
   // sends back one event or errors
   const id = req.params.id
-  const findEventById = Event.findById(id).populate('placeId')
-
-  errorIfNotFound(findEventById, errMessages.GET_ID_NOT_FOUND)
+  Event.findById(id)
+    .populate('placeId')
+    .then(rejectIfEmpty(errMessages.GET_ID_NOT_FOUND))
     .then(event => res.status(200).send(event))
     .catch(next)
 }
@@ -38,9 +38,8 @@ eventController.update = (req, res, next) => {
   // receives updated json for event in body
   // updates or errors
   const id = req.params.id
-  const findEventByIdAndUpdate = Event.findByIdAndUpdate(id, req.body, { new: true })
-  
-  errorIfNotFound(findEventByIdAndUpdate, errMessages.UPDATE_ID_NOT_FOUND)
+  Event.findByIdAndUpdate(id, req.body, { new: true })
+    .then(rejectIfEmpty(errMessages.UPDATE_ID_NOT_FOUND))
     .then(updatedEvent => res.status(200).send(updatedEvent))
     .catch(next)
 }
@@ -50,7 +49,8 @@ eventController.delete = (req, res, next) => {
   // deletes or errors
   const id = req.params.id
 
-  errorIfNotFound(Event.findByIdAndRemove(id), errMessages.DELETE_ID_NOT_FOUND)
+  Event.findByIdAndRemove(id)
+    .then(rejectIfEmpty(errMessages.DELETE_ID_NOT_FOUND))
     .then(() => res.status(204).send())
     .catch(next)
 }
--- a/src/controllers/event.js
+++ b/src/controllers/event.js
@@ -1,5 +1,5 @@
 const Event = require('../models/Event')
-const { rejectIfEmpty } = require('../db/utils')
+const { rejectIfNull } = require('../db/utils')
 const { messages: errMessages } = require('../constants/errors')
 
 const eventController = module.exports = {}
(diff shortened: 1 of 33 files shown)


mattlub:
@des-des @m4v15 

I updated the getAll function [as so](https://github.com/foundersandcoders/open-tourism-platform/blob/f5b3b2274e707116826d16cc048846d28a0c6f86/src/controllers/event.js#L7-L12) and it works nicely.

However, I cannot add `.populate` to the custom `getByIdOrError` function (defined [here](https://github.com/foundersandcoders/open-tourism-platform/blob/f5b3b2274e707116826d16cc048846d28a0c6f86/src/db/utils.js#L10-L18)). I kind of understand why it shouldn't, it's no longer a mongoose `query` so you can't do `.populate`, but I'm not certain about lots of promises stuff.

Is there any solution you can see, the only ideas I have are:
- pass an options object to the `getByIdOrError` function, including populate options (quite messy)
- somehow construct and return a mongoose query instead of a resolved/rejected promise (difficult)

m4v15:
@mattlub When you say you can't add it to the getByIdOrError function, do you mean you can't chain .populate after it, or you can't literally put it inside where we define the function, at the end of [this line](https://github.com/foundersandcoders/open-tourism-platform/blob/f5b3b2274e707116826d16cc048846d28a0c6f86/src/db/utils.js#L11)?

mattlub:
@m4v15 can't chain it on `getByIdOrError`. Don't want to put it inside because the function is generic and used for all the models.

m4v15:
Ah I see - I think passing a populate options then wouldn't be super messy eg:
```
const findByIdOrError = function (id, fieldsToPopulate) {
  return this.findById(id)
    .populate(fieldsToPopulate||'')
    .then(res => {
      if (res === null) {
        return Promise.reject(createCustomDbError(errMessages.GET_ID_NOT_FOUND))
      }
      return Promise.resolve(res)
    })
}
```

If you pass populate an empty string it won't do anything (if you pass it null it will crash)

I mean I guess this option isn't ideal, as we're running populate when we don't need to, but I think it will work

mattlub:
@m4v15 @des-des 
```js
const findByIdOrError = function (id, options) {
  const query = this.findById(id)
  if (options && options.populate) {
    query.populate(options.populate)
  }
  return query 
    .then(res => {
      if (res === null) {
        return Promise.reject(createCustomDbError(errMessages.GET_ID_NOT_FOUND))
      }
      return Promise.resolve(res)
    })
}
```

not overly messy, but in the controller it's slightly inconsistent because you would have:
```js
Event.find()
  .populate('place')
  .then(etc.)
```
and
```js
Event.findByIdOrError(id, {populate: 'place'})
  .then(etc.)
```

thoughts?

m4v15:
@mattlub  Yeah or that, which is nicer (you could put the if on one line?) - or are you thinking this is too messy

m4v15:
@mattlub, sorry I never saw your edited comment before. Yeah I mean I can't see another way of doing it really - the whole point of not doing a mongoose query was so we could get an error - the inconsistency is slightly annoying but not the end of the world?

Also I just added a dropCollectionsAndEnd function to tidy up your test - there might be a better way of doing it but recursion and promises was wigging me out a little and this way makes sense to me (somewhat).

des-des:
I think better just to have a wrapper


```js
rejectIfEmpty = dbRes => dbRes.length === 0
  ? Promise.reject(notFoundError())
  : dbRes
}
```

```js
(req, res, next) => {
  rejectIfEmpty(Events.findOne({}))
    .then(event => res.json(res))
    .catch(next)
 }
```
Then you can handle the error in next

des-des:
But whatever works for you

mattlub:
@des-des  @m4v15 Have done it more like @des-des 's way, I think it's better as it removes the need for the custom mongoose model methods (`getByIdOrError` etc.). Have a look, if you approve I will implement it for all the models.

(look at the files changed instead of the individual commits because I changed things a lot in the last commit)

src/controllers/event.js
```diff
@@ -17,9 +17,9 @@ eventController.getById = (req, res, next) => {
   // receives id in url
   // sends back one event or errors
   const id = req.params.id
-  const findEventById = Event.findById(id).populate('placeId')
-
-  errorIfNotFound(findEventById, errMessages.GET_ID_NOT_FOUND)
+  Event.findById(id)
+    .populate('placeId')
+    .then(rejectIfEmpty(errMessages.GET_ID_NOT_FOUND))
```
des-des: oh yum :ok_hand:

des-des (changes requested):
Okay, I only have one suggestion, we call the function `rejectIfNull` as that is more accurate.

mattlub:
@des-des @m4v15 updated and ready for review

m4v15 (approved):

des-des (approved):

PR #95, opened 2017-08-03
/ Sessions / Login / Signup

des-des:
Update from Mavis
---
relates #100 
I think the functionality of this is done now. This allows us to secure routes and allow users with correct tokens through routes. No routes have actually been secured as of yet

- Upon registering, a password is hashed into the db
- Upon login, a token is sent back in a cookie
- Routes can be secured by checking for a JWT in a cookie
- Routes can be permissioned by checking the scope of the user against a minRole allowed on the route
- There are tests for all the functionality (I think)
- also adds and tests new boomError handler

### Questions

 - When are we actually going to secure the API routes? That has not been done here.
 - We could hash the password as a pre-save function on the User Schema, something like this:
```js
// Execute before each user.save() call
UserSchema.pre('save', function(callback) {
  var user = this;

  // Break out if the password hasn't changed
  if (!user.isModified('password')) return callback();

  // Password changed so we need to hash it
  bcrypt.genSalt(5, function(err, salt) {
    if (err) return callback(err);

    bcrypt.hash(user.password, salt, null, function(err, hash) {
      if (err) return callback(err);
      user.password = hash;
      callback();
    });
  });
});
```
Not sure if we want to but would mean we couldn't ever add users without a hashed password

Old
---
 * Sessions, we need sessions for users logged directly into OPT. Right now this pull implements JWT based sessions. Currently the auth expects the the JWT to be in a header, we should move this into a cookie so that it does not conflict with the oauth. This session / auth will be used for users who want to do 2 things.
   1. Create new apps
   2. authorise apps
 * Permissions: Create middleware that can help us permission our endpoint.
   1. First we have authenticate the user, either from oauth token or from JWT in cookie.
   2. Second, on endpoints that require permissioning, we check the users credentials against those required for their action.


mattlub:
@des-des please could you sum up what you're doing here (bullets maybe), and how it fits in with the other PR? Will be easier to pick up.

e.g. the other one needs a function which takes a request and returns a user (see [here](https://github.com/foundersandcoders/open-tourism-platform/pull/92/files#diff-b7e9fce0c73de87ec22ce5f59ef7668cR35)).

des-des:
@mattlub 
The authorisation step (issueing a code) need the user to be authenticated. The login / session management in this pull will allow us to log a user into the platform and check the users credentials when the try to authorise an application.

I have also updated the top level comment

m4v15:
So do we need to change [this bit](https://github.com/foundersandcoders/open-tourism-platform/pull/95/files#diff-004d1e1bf2f9641418397529818e0943R64) to put the stuff into a cookie?  (at the moment, is this what allows you to use `req.user` in authUser.js)

des-des:
@m4v15 yes, but also here https://github.com/foundersandcoders/open-tourism-platform/pull/95/files#diff-ba775ae7fd1ea884755adaf5125dc306R3

src/routers/apiRouter.js
```diff
@@ -41,4 +47,18 @@ router.route('/products/:id')
   .put(productController.update)
   .delete(productController.delete)
 
+router.route('/sessions/login')
```
mattlub: these should be in the authRouter right?

src/middleware/authUser.js
```diff
@@ -0,0 +1,32 @@
+const boom = require('express-boom')
+
+const roles = require('../constants/roles.js')
+const User = require('../models/User')
+
+const verifyRole = ({ minRole }) => user => {
+  const userScope = user.scope
```
mattlub: why user scope, not user role?

src/controllers/session.js
```diff
@@ -0,0 +1,68 @@
...
+    if (user.length === 0) {
+      return bcrypt.hash(req.body.password, 12)
+    }
+
+    res.boom.badRequest('username already exists')
+  }).then(passwordHash => {
+    const { englishName, arabicName, email, username, imageUrl } = req.body
+    const en = englishName && { name: englishName }
+    const ar = arabicName && { name: arabicName }
+    return Users.create(Object.assign(
```
mattlub: why using .assign instead of just 
```js
{
 username,
 etc...,
 en,
 ar
}
```
?

src/controllers/session.js
```diff
@@ -0,0 +1,68 @@
...
+      { ar },
+      { en }
+    ))
+  }).then(user => {
+    res.send('registered!')
+  }).catch(next)
+}
+
+sessionController.login = (req, res, next) => {
+  Users.find({ username: req.body.username }).then(users => {
```
mattlub: `findOne`

mattlub (changes requested):

src/controllers/session.js
m4v15: I'm assuming its to ensure `en` and `ar` are not mutated

mattlub: but we have just set them. maybe if they are undefined they're not set or something

mattlub: dunno

mattlub: @des-des

mattlub: nah it still sets them as undefined

m4v15:
I've changed it to use cookies and now test the routes, just need to add tests for the middleware

src/middleware/authUser.js
```diff
@@ -0,0 +1,36 @@
...
+      return Promise.reject(boom.unauthorized(auth.UNAUTHORIZED))
+    }
+
+    req.user.id = users[0].id
+    return req.user
+  })
+}
+
+module.exports = opts => (req, res, next) => {
+  addUserIdToSession(req)
```
m4v15: @des-des just getting things straight

ATM, we add their ID to the session, then check their role is allowed on this route, then let them through.

Does it not make more sense to check their role is allowed before adding their id? Not that it makes much difference but it seems weird this way.

Also, why do we not just add their id into the JWT, along with their username [here](https://github.com/foundersandcoders/open-tourism-platform/pull/95/files#diff-004d1e1bf2f9641418397529818e0943R50). I feel like we talked about this but now I can't remember.

des-des: @m4v15 

  1. We can add the username to the jwt.
  2. We need the id to check the permission, ie `PUT events/idofeoinsevent`, that need the id.
  3. We want the same user/session/auth object attached to the request, in both the oauth and the cookie/jwt case, its not too important how we get there

This is the flow I see

Request -> auth -> permission -> controller
auth: verifies the identity of the user, ie checks cookie / auth header and attaches user object to request (id, role, username ..)
permission: optional, if present asserts on the user object. ie: must be logged in; or, must have role SUPER; or must have same id as owner of requested resource

Does that make sense / bring clarity?

m4v15: @des-des  kind of although a couple of things:

> 1. We can add the username to the jwt.

We already do don't we? I was asking why we don't add the id to the JWT at the same time, instead of adding it later during the 'auth' stage

Also, with regards to your flow, what is authUser.js doing? it's both attaching user objects to the request (the auth part) and then verifying the role (i.e. doing some sort of permissioning). But then the authSession.js also does part of your auth (authenticates token, attaches what's in it to the request)

The specific flow we implemented:
1. Make a login POST request, response contains token in cookie, which contains the username and user role
2. Attempt to access secure route with token in cookie
3. authSession.js authenticates token, and puts token into req.user 
4. authUser.js checks user exists, and adds user ID to request
5. authUser.js verifies role and allows or disallows

Is authUser in this instance just an example of the auth/permission layers in your flow?

m4v15: I think basically I'm not sure whether authUser.js is middleware we would use everywhere, or just an example of some middleware we might use.

src/controllers/session.js
m4v15: we have just set them but it's about making sure that the new object has it's own things that are `ar` or `en` and don't point to other things in memory that are potentially mutatable.

I think. @des-des ?

m4v15:
@des-des @mattlub I think this is ready for an initial review, finishing off #100

src/controllers/session.js
mattlub: I don't think it's necessary

.travis.yml
```diff
@@ -4,6 +4,6 @@ node_js:
 services:
   - mongodb
 env:
-  - MONGODB_URI=mongodb://127.0.0.1/
+  - MONGODB_URI=mongodb://127.0.0.1/ JWT_SECRET=somesecret
```
mattlub, line 7: should this be here?

src/controllers/session.js
```diff
@@ -0,0 +1,67 @@
...
+        password: passwordHash,
+        role: roles.BASIC,
+        email: email,
+        imageUrl: imageUrl
+      },
+      { ar },
+      { en }
+    ))
+  }).then(user => {
+    res.send('registered!')
```
mattlub: what do we want to do here?
redirect?

src/controllers/session.js
```diff
@@ -0,0 +1,67 @@
...
+        (err, token) => {
+          if (err) return reject(err)
+
+          resolve(token)
+        }
+      )
+    })
+  }).then(token => {
+    res.cookie('token', token)
+    res.send('success')
```
mattlub: again, do we want to redirect here?

src/controllers/session.js
des-des: hmm I expect I am just being weird

des-des: I expect I am just being weird

src/controllers/session.js
```diff
@@ -0,0 +1,67 @@
...
+
+const sessionController = module.exports = {}
+
+sessionController.register = (req, res, next) => {
+  Users.find({ username: req.body.username }).then(user => {
+    if (user.length === 0) {
+      return bcrypt.hash(req.body.password, 12)
+    }
+
+    res.boom.badRequest('username already exists')
```
des-des: okay, so I actually decided to go this way:

https://github.com/foundersandcoders/open-tourism-platform/pull/95/files#diff-5fc88330781b3266fd9620ce730a474eR29

Ie, reject the promise chain with a boom error. Then it gets passed to next in the catch statement, finally I added middleware that handles this, still not sure what approach is better we should prob choose.

des-des: yeah I think second way is better

src/controllers/session.js
```diff
@@ -0,0 +1,67 @@
...
+  }).catch(next)
+}
+
+sessionController.login = (req, res, next) => {
+  Users.findOne({ username: req.body.username }).then(existingUser => {
+    if (existingUser) {
+      return bcrypt.compare(req.body.password, existingUser.password).then(match => {
+        if (match) return existingUser
+
+        res.boom.badRequest('username / password combination do not exist')
```
des-des: Yeah the same, I think next(boom ...

src/controllers/session.js
```diff
@@ -0,0 +1,67 @@
...
+        },
+        (err, token) => {
+          if (err) return reject(err)
+
+          resolve(token)
+        }
+      )
+    })
+  }).then(token => {
+    res.cookie('token', token)
```
des-des: Set `isSecure` and `httpOnly` to `process.env.NODE_ENV==='production'`

http://expressjs.com/en/api.html#res.cookie

.travis.yml
m4v15: Yeah so we have this JWT_SECRET variable for signing the JWT's and stuff, Travis needs it for the tests. As it's just going to be on the test database I don't think we need to encrypt this, but we can.

src/controllers/session.js
m4v15: removed

src/middleware/authHelpers.js
```diff
@@ -0,0 +1,43 @@
...
+const roles = require('../constants/roles.js')
+const { auth } = require('../constants/errors.json')
+const boom = require('boom')
+
+module.exports = {
+  // Function to get the token, which will be decoded and used to populate req.user
+  // atm it will check the headers, then query params, then cookies
+  // We might need to get rid of the headers check if we end up putting the OAth token their?
+  getToken: (req) => {
+    if (req.headers.authorization && req.headers.authorization.split(' ')[0] === 'Bearer') {
```
des-des: I expect express-jwt will throw an error for the oauth stuff with this.

I think for simplicities sake jus look in the cookie

src/middleware/authHelpers.js
```diff
@@ -0,0 +1,43 @@
+const User = require('../models/User')
```
des-des: maybe put this file in `src/utils/auth_middleware_helpers.js`

Not to bothered either way

src/middleware/authUser.js
```diff
@@ -0,0 +1,8 @@
+const { verifyRole, addUserIdToSession } = require('./authHelpers.js')
```
des-des: I am not sure moving these function out of this file is that useful.

They are the functionality that makes this function.

You could export like
```js
module.exports.addUserIdToSession = addUserIdToSession
```
if you want to test.

But I think the idea is that this file has the code that makes up this middleware. Thoughts?

src/server.js
```diff
@@ -5,25 +5,30 @@ const bodyParser = require('body-parser')
...
 server.use(cors())
 server.use(express.static(path.join(__dirname, 'public')))
 server.use(bodyParser.urlencoded({ extended: true }))
 server.use(bodyParser.json({ extended: true }))
 server.use(boom())
+server.use(cookieParser())
 
 server.use(apiRouter)
-server.use('/oauth', authRouter)
+server.use(authRouter)
```
des-des: can we make it like
```js
server.use('/auth', authRouter)
```

That means login + register + all the oauth stuff will be under `/auth` I think thats good

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,114 @@
...
+tape('POST /register with new user', t => {
+  supertest(server)
+    .post('/register')
+    .send(regUser)
+    .expect(200)
+    .expect('Content-Type', /text/)
+    .end((err, res) => {
+      if (err) t.fail(err)
+      t.equal(res.text, 'registered!', 'should return \'registered!\'')
+      dropCollectionAndEnd(User, t)
```
des-des: I think you need to check the user is in the users table?

des-des: Also, I think register should also log the user in?

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,114 @@
...
+    })
+})
+
+tape('POST /register with new user and then log in', t => {
+  // register user before logging in as that user
+  supertest(server)
+    .post('/register')
```
des-des: okay this is not a biggie, but I think unit test would be better.

Its a bit of a faff, but you could add a user fixture with a salted password in it, then unit test the login endpoint.

You want wanted to go further you could then decode the jwt..

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,114 @@
...
+})
+
+tape('GET SUPER secure route /app with authorized user', t => {
+  // register new user
+  supertest(server)
+    .post('/register')
+    .send(regUser)
+    .expect(200)
+    .expect('Content-Type', /text/)
+    .end((err, registeredRes) => {
```
des-des: image

https://github.com/visionmedia/supertest

That will allow us to flatten stuff.

There is lots of noise here,

what would be nicer
```js
test('GET SUPER secure route /app with authorized user', t => {
   initDb(mockJson)
    .then(() => {
      return supertest(server)
        .get('/apps')
        .set('Cookie', loggedInRes.header['set-cookie'][0])
        .expect(200)
        .expect('Content-Type', /text/)
    }).then(secureRes => {
      t.equal(secureRes.text, 'IN!', '\'IN!\' sent back as a response')
      dropCollectionAndEnd(User, t)
    })
}

```
I am kinda leaving leaving how far you want to take all this up to you, its not a blocker.

des-des (commented):
@m4v15 lol cant request changes on my own pull request (kind of fair enough)

tests/controllers/sessions.test.js
des-des: But yeah, i think if we wanted to make these tests better we would do 2 things:
  1. Move toward testing endpoints in isolation, with more intelligent fixtures.
  2. Testing you results more properly. (Password is in db and hashed, jwt is properly signed, bad jwt fails ...)

.travis.yml
des-des: yeah this is fine, as long as it is not the same as the one we are using in production. (that one should be long and random)

src/middleware/authUser.js
m4v15: Yeah I thought after taking it out it then seemed a bit pointless/unreadable, but then it was mostly to test, I'll do it this way instead

src/controllers/session.js
m4v15: done

src/controllers/session.js
m4v15: @des-des  
as in literally
```js
const inProduction = process.env.NODE_ENV === 'production'
res.cookie('token', token, { isSecure: inProduction, httpOnly: inProduction }
```
?

src/middleware/authHelpers.js
m4v15: yeah I thought that as I wrote the comment, will do

tests/controllers/sessions.test.js
m4v15: I understand your point and definitely with the tests need to be way more comprehensive but I don't really understand your first comment here

m4v15: or maybe I do thinking about it

src/controllers/session.js
des-des: @m4v15 yup seems good

m4v15:
reminder for me of what to do after talking to eoin a bit:

- [x] use promises with super test
- [x] more thorough tests
- [x] make verify role a pure function
- [x] add an allowed through thing

src/controllers/session.js
m4v15: Yeah I mean this was just a place holder for now, I guess yeah a redirect makes sense, not sure what to as atm we don't have any views to redirect to. Maybe `GET /apps` when it exists? But leave it like this for this pull? I can add a comment

src/server.js
mattlub: i had changed it to have `/login`, `/register`
and `/oauth/authorize`, `oauth/token`

are you suggesting `auth/oauth/authorize` or `auth/authorize`?

mattlub:
I haven't really looked properly at all of this- but does it include a function which takes a request and returns a user? This is what the oauth module needs

m4v15:
@mattlub 

So authSession.js looks at a request, looks for a token in a cookie, then verifies the token with our secret and attaches anything in the token onto req.user (at the moment we put username and user role in the token)

Is this what you mean?

tests/helpers/index.js
```diff
@@ -35,3 +39,15 @@ helpers.spyGeneratorErrorMiddlewareCaller = errorMiddleware => (error) => {
   errorMiddleware(error, undefined, resSpy, nextSpy)
   return { resSpy, nextSpy }
 }
+
+helpers.addUserWithHashedPassword = validUser1 => {
+  return new Promise((resolve, reject) => {
+    bcrypt.hash(validUser1.password, 12)
+    .then(hashedPassword => {
+      User.create(Object.assign({}, validUser1, {password: hashedPassword}))
```
m4v15, line 47: flatten promise

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,149 @@
...
+    .get('/apps')
+    .set('Cookie', `token=${token}`)
+    .expect(401)
+    .expect('Content-Type', /json/)
+    .then(res => {
+      t.equal(res.body.error, 'Unauthorized', 'Unauthorised error returned')
+      t.equal(res.body.message, auth.UNAUTHORIZED, `'${auth.UNAUTHORIZED}' is returned as error message`)
+      dropCollectionAndEnd(User, t)
+    })
+    .catch(err => t.end(err))
```
m4v15: flatten promise

m4v15: in all the tests here

tests/middleware/authHelpers.test.js
```diff
@@ -0,0 +1,113 @@
...
+    user: {
+      username: validUser1.username
+    }
+  }
+  User.create(validUser1)
+    .then(() => {
+      return authenticateUserAndAddId(req)
+    })
+    .then(result => {
+      t.ok(result, 'result does exist')
```
m4v15: make the test here check right result is passed, right id added

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,149 @@
...
+  .then(() => {
+    supertest(server)
+    .post('/login')
+    .send({ username: validUser1.username, password: validUser1.password })
+    .expect(200)
+    .expect('Content-Type', /text/)
+    .then(res => {
+      t.equal(res.text, 'success', 'should return \'success\'')
+      t.ok(res.headers['set-cookie'], 'set cookie header exists')
+      t.ok(res.headers['set-cookie'][0].includes('token'), 'Cookie header contains token')
```
m4v15: decrypt the token here and to check it's right maybe?

m4v15: DONE this

tests/helpers/index.test.js
```diff
@@ -0,0 +1,19 @@
+const { validUser1 } = require('../fixtures/users.json')
+const helpers = require('../helpers/index.js')
```
des-des, line 2: OK ... Here you are testing the test helpers, not the application?

tests/index.test.js
```diff
@@ -17,6 +17,8 @@ dbConnection.once('open', () => {
     .then(() => t.end())
     .catch(err => t.end(err))
   })
+  // test the test helper functions
```
des-des, line 20: okay cool

src/controllers/session.js
```diff
@@ -0,0 +1,101 @@
...
+  })
+}
+
+const setTokenCookie = (res, token) => {
+  res.cookie('token', token, {isSecure: isProduction, httpOnly: isProduction})
+}
+
+const registerNewUser = (data) => {
+  if (data.password !== data['confirm-password']) {
+    return Promise.reject(boom.badRequest(authErr.PASSWORDSDONOTMATCH))
```
des-des, line 36: :+1:

src/middleware/authenticateUser.js
```diff
@@ -0,0 +1,24 @@
+const User = require('../models/User')
+const { auth } = require('../constants/errors.json')
+const boom = require('boom')
+
+const authenticateUserAndAddId = (req) => {
```
des-des: This function does not authenticate.. I think this is mostly a naming problem?

src/middleware/authenticateUser.js
```diff
@@ -0,0 +1,24 @@
...
+        return Promise.reject(boom.unauthorized(auth.UNAUTHORIZED))
+      }
+
+      req.user.id = user.id
+      // user is returned for use in oauth
+      return req.user
+    })
+}
+
+module.exports = (req, res, next) => {
```
des-des: Yeah, there is no real authentication happening in this file, all we do is check to see if there is a valid username on the request.user object.

I expect the logic overall is fine, but naming is problematic

src/middleware/rolePermission.js
```diff
@@ -0,0 +1,20 @@
+const roles = require('../constants/roles.js')
+const { auth } = require('../constants/errors.json')
+const boom = require('boom')
+
+// Recieves req.user from addUserIdToSession, and we extract the user's role from that
+const rolePermissionIsSufficient = ({ minRole }) => user => {
+  const rolesOrder = [roles.BASIC, roles.ADMIN, roles.SUPER]
+
+  return (rolesOrder.indexOf(user.role) >= rolesOrder.indexOf(minRole))
```
des-des: Yes, this is so much nicer!

des-des: One problem that I see is that `-1 === -1`. This is not a big issue but you could

```js
const rolePermissionIsSufficient = ({ minRole }) => {
  const rolesOrder = [roles.BASIC, roles.ADMIN, roles.SUPER]
  if (rolesOrder.indexOf(minRole) === -1) {
    throw new Error(..) // this will throw during startup so no problem
  }
  return user => {
    ...
  }
}
```

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,148 @@
...
+  )
+  .then(res => {
+    t.equal(res.body.message, auth.WRONGUSERORPW, 'Correct error message returned')
+    dropCollectionAndEnd(User, t)
+  })
+  .catch(err => t.end(err))
+})
+
+tape('POST /apps with validToken', t => {
+  addUserWithHashedPassword(validUser1)
```
des-des, line 99: and here

tests/controllers/sessions.test.js
```diff
@@ -0,0 +1,148 @@
...
+
+tape('POST /login with validUser1', t => {
+  addUserWithHashedPassword(validUser1)
+  .then(() => {
+    supertest(server)
+    .post('/login')
+    .send({ username: validUser1.username, password: validUser1.password })
+    .expect(200)
+    .expect('Content-Type', /text/)
+    .then(res => {
```
des-des: This is one level too deep?

```js
  doStuff().then(() => {
    return supertest(server)
      .post(..)
      .assert()
    }).then(() => {
      // tests here, without unnecessary promise nesting
    })

des-des (commented):
Okay, cannot request changes.. but see above

src/middleware/authenticateUser.js
m4v15: Yeah I mean we debated this name alot tbh. Isn't checking to see if a username is valid authenticating it? Verify instead of authenticate? What do you think of as authenticating?

des-des: Checking the JWT / oauth token are valid is the step I would consider authentication

tests/controllers/sessions.test.js
m4v15: Why here? I get the other one above (and have fixed) but here supertest is already being returned

m4v15:
@des-des changed name to validate, and fixed the other things. I really hope this is done now...

des-des:
Woop

Issue #104, opened 2017-08-15
add views relating to auth and OAuth


mattlub:
high priority:
- [x] login (`GET /login`)
- [x] register (`GET /register`)
  - check the `User` model for required fields 
- [x] authorise app (`GET /oauth/authorize`)
  - query params `client_id`, `redirect_uri`, `state` (iirc- check them)
  - renders page specific to the client (app)

lower priority:
- [ ] register app (`GET /oauth/register-app`)
  - requires authentication
  - check the `Client` model for required fields 
- [ ] my apps 
  - requires authentication

@des-des @m4v15 thoughts on the endpoints I've suggested?

m4v15:
@JWLD  @mattlub Just wanted to make you aware that currently, the post request to /register expects a particular kind of body, like this:
```js
    const { englishName, arabicName, email, username, imageUrl } = req.body
    const en = englishName && { name: englishName }
    const ar = arabicName && { name: arabicName }
    return Users.create(Object.assign(
      {
        username: username,
        password: passwordHash,
        role: roles.BASIC,
        email: email,
        imageUrl: imageUrl
      },
      { ar },
      { en }
    ))
```

we can allow for organisation name as englishOrgNam and arabicOrgName (didn't do it originally but am going to do it before the PR is finished

m4v15:
Those endpoints look good to me - you know what each form is POST ing to as well yeah?

m4v15:
Are we going to get rid of the user endpoints on the API?

Or I guess we always said we'd just secure these for only `SUPER`

des-des:
@mattlub they look good.

des-des:
scope done for this sprint

PR #106, opened 2017-08-15
Login view

mattlub:
relates #104 
- [x] sets up express handlebars
- [x] login view
- [x] use tachyons

**UPDATE: READY FOR REVIEW**


mattlub:
@JWLD should be picking this up

mattlub:
... I did it in the end... "styling" is super basic for now (but super reusable)

can be changed later lol

screen shot 2017-08-16 at 15 48 29

m4v15 (commented):
@mattlub @des-des Looks great tbh, but a couple of things:

Do we need to make to dual language, cos it's not atm?

If we were planning on using the same style as Data Entry app we're not atm since I think they're planning on using BEM css (I cut out when we were talking about this the other day, sorry if I should have raised it then) @JWLD can confirm.

mattlub:
@m4v15 
- dual language, I think so, but could be added after (what methods could we use to do it?).
- Jack did say Reem was planning to use BEM, but I figure we can change this after the data entry is done if we need to, or the data entry can just use these styles

Basically my idea was to get something ok merged since we are on a tight timescale.

m4v15:
@mattlub 

Yeah sounds good, I just was making sure we were aware of those things.

m4v15 (approved):

src/tachyonsClasses.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
des-des: If you want to reuse you tachyons classes, I would use a handlebars partial.

I think this is fine for now as styling will go away. Could we move this file out of root  though. `src/views` or something

des-des (changes requested):

mattlub:
@des-des I've made the changes :shipit:

des-des (approved):

PR #108, opened 2017-08-16
Authorize app view

mattlub:
relates #104 
authorize app view.

- redirects to login with 'return_to' query params if not logged in
- renders authorize page if logged in


m4v15 (approved):

des-des:
@mattlub reassign me so I know you are not still working on this thnx x

tests/oauth/authorize.test.js
```diff
@@ -0,0 +1,91 @@
...
+      client_id: createdClient.id,
+      redirect_uri: createdClient.redirectUris[0],
+      state: 'random'
+    })
+    .expect(302)
+    .expect('Location', /login/)
+    .expect('Location', /return_to/)
+  .then(res => {
+    const parsedLocationUrl = url.parse(res.headers.location)
+    const locationQueries = qs.parse(parsedLocationUrl.query)
```
m4v15: were you going to test these?

m4v15, line 68: Nothing happens with these variables - we should test them?

m4v15 (changes requested):

m4v15 (approved):

src/controllers/oauth.js
```diff
@@ -11,19 +17,35 @@ const oauth = new OAuthServer({ model: authModel })
...
 oauthController.getAuthorizePage = (req, res, next) => {
-  // get query params to find out what the app is (client_id is required)
   if (!req.query || !req.query.client_id) {
-    return res.boom.badRequest('no client_id provided')
+    return res.boom.badRequest(errMessages.NO_CLIENT_ID)
+  }
+  if (!req.query.redirect_uri) {
+    return res.boom.badRequest(errMessages.NO_REDIRECT_URI)
+  }
+  if (!req.query.state) {
```
des-des, line 26: Okay, I have been hitting my head against this trying to work out if we need it.

One the one hand, it seems like a pretty key security measure (http://www.thread-safe.com/2014/05/the-correct-use-of-state-parameter-in.html), on the other, github does not set it to be a required field. Not really sure what to take from that..

I have not used this parameter when building a client with gh oauth in the past, I expect it adds another level of complexity for the implementer, if it is done properly.

I think we should prob configure auth2-server to allow empty state: https://github.com/oauthjs/node-oauth2-server/blob/master/lib/handlers/authorize-handler.js#L62, and drop the constraint here.

Thoughts?

src/controllers/oauth.js
```diff
@@ -11,19 +17,35 @@ const oauth = new OAuthServer({ model: authModel })
...
+    return res.boom.badRequest(errMessages.NO_STATE)
   }
-  Client.findOne({ id: req.query.client_id })
+
+  if (!req.user) {
+    const queries = {
+      client_id: req.query.client_id,
```
des-des, line 32: What about `redirect_uri` and `state`?

des-des (changes requested):

des-des:
@mattlub heya, sorry this took me so long!

src/controllers/oauth.js
m4v15: @des-des It's all contained in the `return_to` variable, including `client id` - it is just included also as a seprate variable in order to use it to get the name of client as well  (see the other PR about login redirect)

src/controllers/oauth.js
mattlub: I think maybe state is only so important in the 'implicit flow' method of OAuth2, i.e. not the 'authorization code' method. So maybe we should remove it.

mattlub: but I don't think it adds that much complexity, so I think we should only remove it if it causes trouble for app developers in the future.

des-des (approved):

PR #112, opened 2017-08-22
add register-app view

mattlub:
relates #104 

"as a user, I want to be able to go to a page to register an app to use OAuth with the OTP"

**assigned low priority as for now, we can manually enter the data entry and events apps into the db**

fields:
- name
- redirectUri
- redirectUri2

### DIFF
--- a/src/routers/authRouter.js
+++ b/src/routers/authRouter.js
@@ -7,6 +7,11 @@ router.route('/login')
     res.render('login')
   })
 
+router.route('/oauth/register-app')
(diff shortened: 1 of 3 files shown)


m4v15:
When is the `POST` endpoint getting made? And it's just going to add new apps to the database right?

mattlub:
@m4v15 I've added priority 3 as this is actually less important, and will need extra work like the POST request handler. We can add apps initially manually.

Issue #121, opened 2017-09-04
Authenticating the access token


m4v15:
How are we securing the routes? 

Should they be accessible only by client's logged in with oauth, or also by people directly logged into the platform. This question affects what we do next.

The module has an authenticate function that returns the token. However this will presumably fail if a request is made that has a JWT in a cookie and no access token?

mattlub:
close this as it was addressed?

mattlub:
#126

PR #122, opened 2017-09-05
add resource owner to permissions

mattlub:
update:

related #134 

- [x] make middleware
  - changes name from `rolePermission` to just `permissions`
  - adds functionality to check if the requesting user owns the resource
- [x] write tests
  - amend old middleware tests to use new syntax
  - write new tests in `permissions.test.js`

**Note:** The middleware has not been properly added to routes yet, when it is, full integration tests for the middleware will be created from the old controller tests
---
the auth is completetly separate from the JWT stuff
WIP: currently failing tests because auth is req'd
relates #121


m4v15 (changes requested):
Get resource owner from the path?

mattlub:
@m4v15 made the change

m4v15 (approved):

src/middleware/permissions.js
```diff
@@ -0,0 +1,75 @@
...
+  const pathName = url.parse(req.url).pathname
+  return routeResourceMapping[pathName.split('/')[1]]
+}
+
+module.exports =
+  ({ minSufficientRole, owningResourceIsSufficient }) => (req, res, next) => {
+    const resourceType = getResourceType(req)
+
+    if (owningResourceIsSufficient && !resourceType) {
+      next(boom.badImplementation())
```
des-des: :ok_hand:

src/middleware/permissions.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if (!req.user || !req.user.id) {
+      return next(boom.unauthorized(auth.UNAUTHORIZED))
+    } else if (hasSufficientRole({ minSufficientRole })(req.user)) {
+      return next()
+    } else if (!owningResourceIsSufficient) {
+      return next(boom.unauthorized(auth.UNAUTHORIZED))
+    } else {
```
des-des: Okay am confused.
hasSufficientRole && (user is trying to modify resources that they do not own)

then we let them through?

this if / else branching not sure about.
```js
if (minRole && !hasSufficientRole(req, minRole)) { Boom.unauthorised }
 
if (needOwner && !hasSufficientOwnership(req) { Boom.unauthorised }
```
Either this is broken, or the naming is incorrect

tests/fixtures/events.json
```diff
@@ -5,5 +5,18 @@
    "invalidEvent1": {},
    "invalidEvent2": { "en":{"name": "Dinner at Alreda"}, "categories": ["eating"] },
    "invalidEvent3": { "en":{"name": "Dinner at Alreda"}, "categories": [null, "dining"] },
-   "invalidEvent4": { "en":{"name": "Dinner at Alreda"}, "categories": []}
+   "invalidEvent4": { "en":{"name": "Dinner at Alreda"}, "categories": []},
+
+   "event": {
```
des-des: why is this called `event`. The ones above are called `validEvent`

tests/fixtures/users.json
```diff
@@ -17,11 +18,11 @@
...
 		}
 	},
-	"regUser":{
-			"username": "m4v15",
-			"password": "hashthis",
-			"confirm-password":"hashthis",
-			"email": "newemail",
-			"name_en":"mavis"
+	"regUser": {
+		"username": "m4v15",
```
des-des: tabs??

tests/middleware/permissions.test.js
```diff
@@ -0,0 +1,55 @@
...
+tape('checkUserOwnsResource with resource \'User\', and the correct user', t => {
+  checkUserOwnsResource(User)(user.id)(user)
+  .then(() => t.end())
+  .catch(err => t.end(err))
+})
+
+tape('checkUserOwnsResource with resource \'User\', and an incorrect user', t => {
+  checkUserOwnsResource(User)(user.id)(wrongUser)
+  .then(() => t.end('promise should reject, as user is wrong'))
+  .catch(() => t.end())
```
des-des, line 43: Having a test without an assertion is kinda weird.
```js
t.ok(true, 'a good thing happened')
```
works

des-des (changes requested):

tests/fixtures/events.json
mattlub: removed, and used validEvent1

tests/fixtures/users.json
mattlub: json formatted

tests/middleware/permissions.test.js
mattlub: I disagree, we run the tests to check there are no failures, it seems pointless to stick an extra `t.pass(...)` or `t.ok(true,...)` in there

src/middleware/permissions.js
mattlub: minSufficientRole is the minimum role which would guarantee you access, regardless of whether the resource owner can also access.

so I did
```js
if (hasSufficientRole) { next }
if (!hasSufficientRole && hasSufficientOwnership) { next }
if (!hasSufficientRole && !hasSufficientOwnership) { boom(unauthorised) }
```

I'm not convinced what I've done will turn out to be the best method, but I'm not entirely sure what you're suggesting.

mattlub:
@des-des I've changed the names now.

src/middleware/permissions.js
```diff
@@ -0,0 +1,84 @@
...
+    if (ownerIsPermitted && !resourceType) {
+      next(boom.badImplementation())
+    }
+
+    if (!req.user || !req.user.id) {
+      return next(boom.unauthorized(auth.UNAUTHORIZED))
+    }
+
+    else if (hasSufficientRole({ minRole })(req.user)) {
+      return next()
```
des-des, line 80: This still makes no sense to me.
This line here says the user is allowed to access the resource, but at this point it is possible that the user is trying to access a resource that they are not allowed, as some of the checks have not been made.

```js
if (userFailsFirstCheck) unautherised
if (userFailsSomeOtherCheck) unautherised
if (somethingElse) unautherised

// authorised!!
```

If you think I am wrong, then ill merge this tomorrow morning.

des-des (changes requested):

src/middleware/permissions.js
m4v15: Which checks haven't been made?

des-des: okay, I see what is going on now (https://github.com/foundersandcoders/open-tourism-platform/issues/134#issuecomment-330338072)

des-des (approved):

des-des:
happy for you to merge

Issue #117, opened 2017-08-29
Finalising V1


m4v15:
There a few things we need to sort out so the V1 of the API can be used effectively (mainly in the events app)

- [x] DB Field naming conventions #96 
- [x] append the api routes with /api/ ? #105 
- [x] date-related query params for getAll events #107 (PR #116)
- [x] finalise the fixed categories #82

Then there are a couple which we should sort out at some point soon but aren't necessarily urgent for the API to be used
 
- [ ] sorting of returned items from GET requests #81 (has been done for events, do we need to worry about other routes now?)
- [ ] don't return '__v' field from requests #97

@des-des Can you meet at 6pm your time?

mattlub:
also finalising other routes, like using `/auth` or `/oauth`

des-des:
@m4v15 Super sorry, I totes missed this, not sure how

mattlub:
also finalising required fields etc.

mattlub:
I would say "append the api routes with /api/ ? #105" needs to be decided, but the other 2 are enhancements and not urgent

mattlub:
closed because v1 was finalised

Issue #96, opened 2017-08-12
DB Field naming conventions


mattlub:
In [this PR](https://github.com/foundersandcoders/open-tourism-platform/pull/94) I am populating the `placeId` field in the `events` model

**I want to change the field name in the events model to `place` instead of `placeId`**
- I would also propose to do the same on all fields which currently end in `'id'`, even if we do not plan to populate them.

The potential problems with this is confusion:
- this would mean on `POST`, you would have to send an `id` under the `place` field

------

@des-des mentioned the possibility of sending it back as `place` when populated, but keeping the db field as `placeId`. The problem here is seeming inconsistency in the model between a `POST` and a `GET`. Does anyone do this?

Thoughts? @m4v15 @des-des @mantagen

mattlub:
I am leaning towards @des-des suggestion now to be honest.

des-des:
Yeah this is a really good question that I do not have an answer to. Maybe @mantagen has some thoughts. I think if you are on the fence, go for the path of least resistance.. ?

m4v15:
Yeah my whole issue with it was that in the DB it's just an id not a place itself, I'm pro calling it place id in the DB models but in the response object call it place. I see what you mean about the `POST` issue, but I don't think it's a massive problem - however you probably have more of an idea of this due doing stuff with Mohamed on the events app.

m4v15:
Although I guess by path of least resistance the easier thing would be to just call it place. Tbh I can also be convinced easily either way

mantagen:
In the DB it's not just an id. It's a populatable reference to an entity, called a _place_.

... I'd probably go with `place`

mattlub:
I think this is agreed, we will remove all appended `id`

Issue #82, opened 2017-07-27
finalise the fixed categories


m4v15:
**Edit by @mattlub**

We are trying to finalise the fixed categories, can people please suggest any they think should be added:
@RQ16 @JWLD @Karyum @Suuuuha @MohamedOmarii @philawsophizing @m4v15 @des-des @mantagen Thanks

[current suggested categories here](https://github.com/foundersandcoders/open-tourism-platform/blob/finalise-db-fields/src/models/constants.json) or below:

### Accessibility options
  - Audio Recordings
  - Braille
  - Big Fonts
  - Carer
  - Place for Guide Dog
  - SMS messaging
  - Sign Language
  - Good Lighting
  - Carer
  - WheelChair Access
  - Disabled Parking
  - Disabled Toilets
  - Carer
### Place Categories
  - cafe
  - music venue
  - religious site
  - food and drink
  - education
  - retail
  - sport
  - municipal
  - healthcare

### Event Categories
  - music
  - dining
  - educational
  - conference
  - sport
  - competition
  - launch
  - party
  - wedding
  - cultural event
  - miscellaneous-

### Product Categories
  - jewellery
  - clothing
  - crochet
  - calligraphy
  - embroidery
  - metalwork
  - food
  - drink
  - pottery
  - glassware
---
**Edit**:
We need to decide on the fixed categories for `roles`, `events`, `products`, `accessibility`.

Can work with current/previous app developers on this

Original
---
We need to finalise these before v1

mattlub:
is this done? docs link to the [constants file](https://github.com/foundersandcoders/open-tourism-platform/blob/master/src/models/constants.json).

Or maybe change the title of the issue to be "finalise the fixed categories" or similar?

UPDATE: mavis updated the title

PR #125, opened 2017-09-06
Finalise api v1

m4v15:
relates #117, specifically:

- removes the appended `Id` from all `___Id` fields #96 
- adds fixed categories #82 
- Also update the docs to have categories explicitly stated

---

I got the accessibility options from v1 of easynaz, events/places from browsing the web/looking at what mohammed has begin to use in naz-events, and products from v1 of Nasijona, plus a couple of other obvious categories I thought of. Let me know if you think we should have others/lose some.

**Note:** when this gets merged to master and pushed to heroku, we will need to reset the heroku database as all the enums have changed.

So because it's mongo we need to wipe the current production database (this may have knock on effects for apps built on the platform that want to see some results, so we may want to think about adding some stuff via Postman when it is live

### DIFF
--- a/data_spec_v1.md
+++ b/data_spec_v1.md
@@ -82,7 +82,7 @@ name | type | required | permissions
 id | KEY | true |-
 createdAt| DATE | - | -
 updatedAt| DATE | - | -
-ownerId | FOREIGN KEY | false | ADMIN
+owner | FOREIGN KEY | false | ADMIN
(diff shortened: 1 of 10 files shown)


mattlub (approved):

des-des (approved):

des-des:
Follow up work required: #128

PR #126, opened 2017-09-07
Validate header token middleware

mattlub:
related #27 

never nexts an error, this should be handled by the permissioning middleware.

- [x] export oauthServer 
- [x] made middleware
  - includes credentialsRequired option
- [x] write tests

### DIFF
--- a/src/controllers/oauth.js
+++ b/src/controllers/oauth.js
@@ -16,6 +16,8 @@ const oauth = new OAuthServer({ model: authModel })
 
 const oauthController = module.exports = {}
 
+oauthController.oauthServer = oauth
+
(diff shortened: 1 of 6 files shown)


m4v15 (approved):

m4v15:
@des-des 

We're thinking now that `validateJWT` and `validateHeader` will let anyone through, whether logged in or not (using `credentials_required=false`) and then we should build a permission layer that just checks whether they are logged in or not, and if so then let them through to the next permission layers (checking their role etc).

All the `validateX` functions will do is populate req.user (or not, if they aren't logged in)

des-des (approved):

Issue #129, opened 2017-09-10
Handle jwt-express errors


m4v15:
atm:

"The default behavior is to throw an error when the token is invalid, so you can add your custom logic to manage unauthorized access as follows:

```javascript
app.use(function (err, req, res, next) {
  if (err.name === 'UnauthorizedError') {
    res.status(401).send('invalid token...');
  }
});
```
"

From the [jwt-express docs](https://github.com/auth0/express-jwt/blob/master/README.md)

If a token is expired we probably want to direct them back to the log in page, or else we should render and nicer looking unauthorized page

mattlub:
@m4v15 can we close?

PR #131, opened 2017-09-10
Remove expired cookie JWT

m4v15:
Solves #129 

Can't think of any other errors we might have to deal with from express JWT but maybe there are, will check

At the moment it checks if its a jwt error, then checks the message and if it's expired it will remove the cookie a redirect back to the original URL (which will deal with the request differently now it has no cookie as opposed to an expired one)


mattlub (changes requested):
can we change the name of the file to JWTErrorHandler or similar?

mattlub:
this seems quite hacky/not ideal but I'm happy to add it in, otherwise we would have to do our own version of the jwt-express module or combine this step with the permissions middleware.

m4v15:
@mattlub addressed and refactored

mattlub (approved):

src/middleware/jwtExpiredErrorHandler.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = (err, req, res, next) => {
+  if (err.name !== errNames.UNAUTHORISED_JWT) {
+    return next(err)
+  }
+
+  switch (err.message) {
+    case errMessages.EXPIRED_JWT:
+      res.clearCookie('token')
+      res.redirect(req.originalUrl)
```
des-des, line 11: Not convinced. I think 403. Ie your middleware has explicitly aborted the request. Either the middleware should let them through, or 403 them. This feels hacky

des-des (changes requested):

src/middleware/jwtExpiredErrorHandler.js
m4v15: Yeah I agree, it was just we have issues with jwt-express.

We're not using it as an authenticater really, we're setting `credentialsRequired:false` so it should let them through whether they have a cookie or not, but should populate req.user if it can.

However, it throws an error if they have an invalid cookie, which seems annoying/gives two different results to 2 very similar circumstances:

- A non logged in user tries to authorize and app, gets redirected to login page
- A user with an expired cookie tries to authorize an app and they get an authorization error

I guess what we want it to do is if the cookie is invalid, it should ignore it and let them through without populating `req.user` (i.e. as though they didn't have a cookie at all) - they will then get redirected to login if it turns out they need to have `req.user` populated

m4v15: I guess for now we can just send a 403 error for now though, but this will be an error that an end user (as opposed to an app/client) will see:

1. User logs in, gets a token.
2. Token expires
3. User goes to an app, app asks them to authorize on OTP, redirects them to `OTP/oauth/authorize` with correct query parameters.
4. This is a secure route, so will check the cookie
5. jwt-express will give the user back an error

mattlub: @des-des it is hacky but we can't change the jwt-express middleware package, and we def don't want to be rejecting requests with expired tokens, we just want to let them through without populating req.user.

solution is either this or scrap the express-jwt package

des-des:
@m4v15 @mattlub I do not see this being a security problem happy to let it through if you are happy with it

mattlub:
@m4v15 I reckon merge it, but raise an issue about rewriting the JWT middleware to be able to remove this middleware

des-des (approved):

PR #136, opened 2017-09-12
Add oauth client get and post routes and tests

m4v15:
Just added handlers so can post and get clients using postman


tests/controllers/oauthClient.test.js
```diff
@@ -0,0 +1,104 @@
...
+tape('emptying db.', t => {
+  Promise.all([
+    User.remove({}),
+    Client.remove({})
+  ])
+  .then(() => t.end())
+  .catch(err => t.end(err))
+})
+
+// Tests for: GET /users
```
mattlub: comment is wrong

tests/controllers/oauthClient.test.js
```diff
@@ -0,0 +1,103 @@
...
+  addUserWithHashedPassword(validUser1)
+  .then(() => makeLoggedInToken(validUser1))
+  .then(token => {
+    return supertest(server)
+    .get('/oauth/clients')
+    .set('Cookie', `token=${token}`)
+    .expect(200)
+  })
+  .then(res => {
+    t.equal(res.body.length, 0, 'should return length 1 array')
```
des-des: Length 1 or 0?

tests/controllers/oauthClient.test.js
```diff
@@ -0,0 +1,103 @@
...
+  addUserWithHashedPassword(validUser1)
+  .then(() => makeLoggedInToken(validUser1))
+  .then(token => {
+    return supertest(server)
+    .get('/oauth/clients')
+    .set('Cookie', `token=${token}`)
+    .expect(200)
+  })
+  .then(res => {
+    t.equal(res.body.length, 1, 'should return length 1 array')
```
des-des: assert on contents more\

tests/controllers/oauthClient.test.js
```diff
@@ -0,0 +1,103 @@
...
+      .send(clientFromFormData)
+      .expect(201)
+    })
+    .then(res => {
+      t.equal(res.body.name, clientFromFormData.name, 'correct name added')
+      t.equal(res.body.user, user.id, 'correct user added')
+      t.equal(res.body.secret.length, 16, '16 digit secret added')
+      t.deepEqual(res.body.redirectUris, clientFromFormData.redirectUris, 'correct redirectUris added')
+      t.deepEqual(res.body.grants, [grants.authCode], 'correct grants added')
+      t.ok(res.body._id, 'client given an id')
```
des-des, line 109: Also look inside the database?

src/controllers/oauthClient.js
```diff
@@ -0,0 +1,29 @@
...
+
+oauthClientController.create = (req, res, next) => {
+  // receives json for user in body
+  // sends back created user
+  const secret = randtoken.generate(16)
+  const newClient = new Client({
+    name: req.body.name,
+    user: req.user.id,
+    secret,
+    grants: [grants.authCode],
```
des-des, line 23: What does this do? (I should defo know that but would appreciate advice if you have worked it out)

des-des (changes requested):

des-des:
@m4v15 also could you add this to docs

src/controllers/oauthClient.js
m4v15: @des-des So the oauth2 has a variety of different flows and grant types and ways to get access tokens - see [here](https://alexbilbie.com/guide-to-oauth-2-grants/)

The auth code flow is the common one that everyone uses (user authorizes, client gets a code, client exchanges code + secret for token), but there are also others like the refresh token grant (user exchanges a refresh token + client_secret for an access token) or implicit grant (which I don't fully understand). Anyway, when we create a client we have to specify the grant types we plan to use with them, in this case, the auth code grant type. I've actually managed to get it to work with a refresh token which we should implement in the future, but we don't need to yet.

tests/controllers/oauthClient.test.js
m4v15: I've done a simple check (findById, check the name is correct) to look inside the database for 2 reasons:

- 1. the first set of tests are just checking what's returned from a mongoose add function effectively, we should be able to be fairly certain that if something is returned and it matches the correct things then it is in the database.
- 2. When I tried to be more thorough I got stuff like this:
```
      operator: deepEqual
      expected: [ 'www.foundersandcoders.com' ]
      actual:   [ 'www.foundersandcoders.com' ]
      at: Client.findById.then.newClient (/Users/mdavis/Desktop/open-tourism-platform/tests/controllers/oauthClient.test.js:114:15)
      stack: |
```
Which is obviously tape just being weird, but as I don't see this as a massively important test (as it's just checking mongoose had worked properly) so I'm not gonna spend ages working out why right now. Also there's a few string conversions and stuff that I really don't think is necessary.

m4v15: Or maybe it's me misunderstanding deepEqual

m4v15:
@des-des @mattlub added to the docs, but it should probably go in the oauth specific docs when #140 is merged. I've left it in just the docs file for now.

des-des (approved):

Issue #133, opened 2017-09-11
links in the docs to constants are wrong in some places


mattlub:
because they link to an old commit.

Issue #103, opened 2017-08-15
Write docs for OAuth flow for clients (apps)


mattlub:
like [these](https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-authorization-options-for-oauth-apps/)

PR #140, opened 2017-09-12
oauth docs

mattlub:
- [x] moves api docs to folder
- [x] fixes category links in api docs #133
- [x] creates oauth flow docs #103 

heavily based on https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/about-authorization-options-for-oauth-apps/


m4v15 (approved):

docs/oauth.md
```diff
@@ -0,0 +1,77 @@
...
+GET https://nazareth-open-tourism-platform.herokuapp.com/oauth/authorize
+```
+
+**Query parameters**
+
+Name | Type | Description
+--- | --- | ---
+client_id | string | **Required.** The client ID you received from Nazareth Open Tourism Platform for your app.
+redirect_uri | string | **Required.** The URL in your application where users will be sent after authorization.
+state | string | **Required.** An unguessable random string. It is used to protect against cross-site request forgery attacks.
```
des-des: Could we get an example of the url with the query params

docs/oauth.md
```diff
@@ -0,0 +1,77 @@
...
+
+### 2. Users are redirected back to your site by Nazareth Open Tourism Platform
+
+If the user accepts your request, Nazareth Open Tourism Platform redirects back to your site with a temporary authorizaion code in a `code` parameter as well as the state you provided in the previous step in a `state` parameter. If the states don't match, the request was created by a third party and the process should be aborted.
```
des-des: > temporary authorizaion code in a `code` parameter

better: temporary authorizaion code in a `code` query parameter

could we have an example a url that the user might be sent to here

des-des: > third party

maybe say malicious

docs/oauth.md
```diff
@@ -0,0 +1,77 @@
...
+
+Exchange the `code` for an access token:
```
des-des: Can you be more descriptive.

Ie once the user is redirected to your server, your server should make a post request over *https* to the url: .

Then add example.

des-des (changes requested):
Okay cool.

Right now you need to talk a bit about actually creating secret, id etc.

Say that they need to get in touch with you to generate this stuff

mattlub:
@des-des added how to get client id etc.
made other changes to improve clarity

m4v15:
Simple example of the oauth flow from the client side here https://github.com/m4v15/OTP_Oauth
 - maybe we can have the app.js file in the docs?

mattlub:
@m4v15 I would say link to it somewhere in our documentation.

des-des (approved):

des-des:
Happy for you too merge this more stuff can be added in another pull if there is more you want to add

m4v15:
@mattlub I fixed the merge conflicts (there weren't any locally for me, I don't really understand what happened). I then changed the GET and POST clients docs to be in the auth docs and secured them so that only SUPER's can post and view clients (for now)

m4v15:
@mattlub cos I've made some changes I've left it for you to merge

mattlub:
@m4v15 should that SUPER user stuff really be in the docs? I was under the impressions super users would literally only be us/the product owner

m4v15:
@mattlub good point, moved

Issue #17, opened 2017-06-30
Event logging


sofer:
We need to make sure that we are logging all user interactions, so that at some point we can interrogate them for a better understanding of how users are interacting with the apps on the platform.

mattlub:
Do you mean logging info about requests to the API and their responses? Or something different/more specific?

sofer:
For now, it really needn't be anything more sophisticated than an http log, with a time stamp, user details (if logged in) and request details.

mattlub:
https://github.com/expressjs/morgan is a popular http logging with express/node package

mattlub:
@des-des @m4v15 should we log to a file/a db collection/other?

e.g. morgan example logging to a new text file each day:
https://github.com/expressjs/morgan#log-file-rotation

des-des:
@mattlub we cannot log directly to fs in heroku, heroku fs is ephemeral.

mattlub:
ok, I don't know where we should go with this then.

- there are logging "add-ons" in Heroku (see [here](https://elements.heroku.com/addons/#logging)) but predictably they don't offer much for free.

- Otherwise you can implement your own ["log drain"](https://devcenter.heroku.com/articles/log-drains) but I don't really know what that means, but must involve 'draining' the logs to somewhere, leaving us with a fair bit to do e.g. host the logs elsewhere

des-des:
@sofer Are there more detailed requirements?

I expect some express logging middleware being hooked up to a mongo table will be fine for MVP @mattlub .

PR #142, opened 2017-09-12
WIP: logging 

mattlub:
- [x] add morgan
- [ ] decide what to do with it (discuss in #17)

@m4v15 @des-des I started a PR


des-des:
If this solution is only for fs we cant do it on heroku

Issue #144, opened 2017-09-13
update routes can overwrite nested objects


m4v15:
We should fix this, easy enough by just converting to object dot notation:

https://github.com/Automattic/mongoose/issues/5285

PR #145, opened 2017-09-13
Object to dot notation

m4v15:
addresses #144

### DIFF
--- a/src/middleware/objectToDotNotation.js
+++ b/src/middleware/objectToDotNotation.js
@@ -1,0 +1,22 @@
+
+const flattenObject = (oldObject) => {
+  const setObject = {}
+  Object.keys(oldObject).forEach((key) => {
+    if (typeof oldObject[key] === 'object') {
+      Object.keys(oldObject[key]).forEach((subkey) => {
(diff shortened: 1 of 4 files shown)


mattlub (approved):

des-des:
@m4v15 I think this might get someone into all kinds of trouble if somewhere down the line we needed nested objects in body..

des-des:
@m4v15 what about this: https://github.com/Automattic/mongoose/issues/5285#issuecomment-309124718

m4v15:
Yeah I did sort of think this, I think we might be better just making this a helper function we can call when we need it, as we might need it when we do field permissions:

- When the req.body comes in for the update it will be a nested object, we might need to flatten this to do our permission field by field check before doing the update in mongoose

I think what I will do is add it as a helper functions, and use that as a pre update and findoneandupdate hook, and then it can be used before the field permission as well if necessary

For now I will put this on hold until #122 and #146 are done

des-des:
I think this should happen final step before mongoose call

m4v15:
hey I'm cleaning my github up, presumably I can close and delete this branch @des-des @Karyum ?

m4v15:
actually I can just unassign myself 😅

Issue #134, opened 2017-09-11
Permissioning


m4v15:
relates #27 

- [x] Add different permission layers:
  - [ ] Logged in or not
  - [ ] Role permission
  - [ ] Owner of Resource Permission
  - [ ] Who is allowed to edit what field permissions
- [x] After the functions are made, we should finalise which routes should be secure and how, and add the appropriate middleware.

mattlub:
for "Who is allowed to edit what field permissions", can we decide if we want to do this and if so, give examples.

mattlub:
new middleware:

```js
fieldPermissions([
      {
        minRole: roles.SUPER,
        field: '_id'
      },
      {
        minRole: roles.ADMIN,
        field: 'owner'
      }
    ])
```

mattlub:
probably better

```js
fieldPermissions({
      _id: roles.SUPER,
      username: roles.ADMIN
    })
```

mattlub:
I've spent/wasted a lot of time on this. I think we need to finalise how we want the middleware to look when implemented on the routes. I still haven't come to a solid conclusion. Thoughts on the following @m4v15 @des-des ?

- `minRole` and `ownerPermitted` would be the first permissions 'barrier'
- then `fields`, if present, would be the second 'barrier'

So we would have:

`PUT`:
```js
permissions({
  minRole: ADMIN, 
  ownerPermitted: true
  fields: {
    _id: { minRole: SUPER, ownerPermitted: false }
    username:  { minRole: ADMIN, ownerPermitted: true }
    name:  { minRole: ADMIN, ownerPermitted: true }
  }
})
```

`DELETE`:
```js
permissions({
  minRole: SUPER
})
```

m4v15:
I think that looks good, and we could have a fixtures file with all of the fields objects for each table to tidy it up:
`PUT`:
```js
permissions({
  minRole: [ADMIN, OWNER],
  fields: UserFieldPermissions
})
```

des-des:
@m4v15 @mattlub 

Hey both approaches seem fine. One thing I would say is that the name `ownerPermitted` does not make lots of sense. It is not so much a question of whether the owner is permitted, but a question of whether a non-owner is permitted?

m4v15:
The owner permitted flag is as a secondary check - so if the route is only for super, But then the owner is permitted if the owner is not super they are still allowed through

des-des:
[...]

PR #146, opened 2017-09-13
Field permissions

mattlub:
**UPDATE: ready for review**

relates #134 

what i've done in this PR:
#### 2) move permissions functions to helpers file `helpers/permissions.js`
  - includes new function `getUnauthorizedFields`
  - includes all other functions used by `permissions.js` and `fieldPermissions.js` middleware

#### 3) adds specific field permissioning middleware
  - it checks if user owns resource
  - then finds any unauthorized fields (using `getUnauthorizedFields`, which just uses `req.user.isResourceOwner`)
  - **I updated this from before (last week)**, as before it was shortcutting and just checking if user owned resource by checking `user.isResourceOwner`, because the `permissions` middleware sets it (but only if it checks it). So this would not work for the case where an admin satisfied the minRole condition, and so resource owner is not checked by `permissions`, but then there were fields with `[ SUPER, OWNER ]` permissions in `fieldPermissions`. The resulting middlewares are admittedly not very pretty and probably could be improved greatly by taking a step back. But I will not be doing this unless I return to Naz :)


mattlub:
UPDATE: failing on Travis but not locally, I don't know why

m4v15:
So it's failing on travis on tests that no longer exist, as far as I can tell

mattlub:
## @m4v15 YESSSS. I fixed the f*cker.

I'm on linux now, and the tests were failing for me but I could debug. The error was that the expiry date had to be set on the access token, so I've filled out the test tokens with all the attributes. God knows why the tests were passing on our machines before (should we be worried about this?) but all good now.

mattlub:
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

m4v15:
Will go through it all tomorrow morning and then hopefully it can get merged - on a side note we now have debug mode on travis if we want it lol

m4v15:
@mattlub So it should be relatively clear but what I've done is change the `checkUserOwnsResource` function to just return `true` or `false` (or a promise that resolves to that) instead of this function doing the actual rejection. This allows us to skip the `.catch(() => {})` in fieldPermissions.js which feels messy to me. Actually more importantly (I think) it also means that the function is now just a helper function that does what the name implies - i.e. check the owner, not reject necessarily. Once it is used, we can then decide whether to reject or not.

I've changed the middlewares to make the check and then reject or not depending on the middleware ([reject for permissions](https://github.com/foundersandcoders/open-tourism-platform/blob/8f049aedcfe5e4fc3feb8c3061cd69fc43d3ed7f/src/middleware/permissions.js#L61-L65) and [just set whether owner or not for fieldpermissions](https://github.com/foundersandcoders/open-tourism-platform/blob/8f049aedcfe5e4fc3feb8c3061cd69fc43d3ed7f/src/middleware/fieldPermissions.js#L51))

I think this is much clearer.

Also I've changed the fieldPermissions to only check for the owner if it is an actual resource (which it won't be for POST) - I'm not 100% sure they way I've done this is best (conditional promise chaining is a bit weird) but I think it's ok ([I resolve a Promise and then use an if statement in a `.then`](https://github.com/foundersandcoders/open-tourism-platform/blob/8f049aedcfe5e4fc3feb8c3061cd69fc43d3ed7f/src/middleware/fieldPermissions.js#L45-L53))

N.B. this is why I thought the PR was broke this morning - sorry, it wasn't really I guess, but the conditional chaining and the `.catch(() => {})` meant error handling was getting wierd

m4v15 (approved):

m4v15:
@mattlub unless any issues, assign @des-des

mattlub:
@m4v15 looks fine

src/helpers/permissions.js
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const checkUserOwnsResource = resourceType => resourceId => user => {
+  /**
+  * Checks Ownership of a resource
+  *
+  * returns a Promise that resolves to true or false, or rejects if
+  * resourceId is not found in DB (404)
+  *
+  * N.B. when the resource type itself is a User, the method is
```
mattlub, line 39: is this really neessary?

src/helpers/permissions.js
```diff
@@ -0,0 +1,115 @@
...
+  * different but the returns are the same
+  */
+
+  if (resourceType === User) {
+    return Promise.resolve(resourceId === user.id.toString())
+  }
+
+  return resourceType.findById(resourceId)
+  .then(rejectIfNull(errMessages.GET_ID_NOT_FOUND))
+  .then(doc => {
```
mattlub: one line?

src/helpers/permissions.js
```diff
@@ -0,0 +1,115 @@
...
+    users: User
+  }
+  const pathName = url.parse(req.url).pathname
+  return routeResourceMapping[pathName.split('/')[1]]
+}
+
+const getUnauthorizedFields = fieldPermissions => fieldsToChange => user => {
+  /*
+  * check an array of the fields a user is trying to change against
+  * an array of those fields and their permission levels
```
mattlub: is this right? what is "an array of those fields and their permission levels"?

src/helpers/permissions.js
```diff
@@ -0,0 +1,115 @@
...
+  * returns an array of fields which the given user is attempting to
+  * change but is not authorised to do so
+  */
+
+  const permissionedFields = Object.keys(fieldPermissions)
+
+  return fieldsToChange
+    // filter out fields which are not permissioned
+    .filter(field => permissionedFields.includes(field))
+      // filter down to fields which user is not permitted to change
```
mattlub: indent

src/helpers/permissions.js
m4v15: yeah this was one line originally then I had a bug and forgot to change it back

src/helpers/permissions.js
m4v15: yeah I guess this is phrased badly

src/helpers/permissions.js
m4v15: @mattlub you wrote this m8

mattlub: You hijacked my PR!

des-des:
Woop, ill have a look at this today

src/middleware/fieldPermissions.js
```diff
@@ -0,0 +1,67 @@
...
+      return next(boom.unauthorized())
+    }
+
+    Promise.resolve().then(() => {
+      // for POST requests, req.params.id will be undefined and we do
+      // not need to check user owns resource anyway
+      if (resourceId) {
+        return checkUserOwnsResource(resourceType)(resourceId)(user)
+        .then(ownerStatus => {
+          req.user.isResourceOwner = ownerStatus
```
des-des, line 51: Hmm this is just a suggestion, but it might be nice to add this in only once, some middleware before this. (right now being added by 2 different middlewares).

But this kinda stuff can wait until we do a refac somewhere down the line..

I expect this might not be a good idea with the way all the middlewares currently compose together.

src/routers/apiRouter.js
```diff
@@ -2,20 +2,36 @@ const userController = require('../controllers/user')
 const placeController = require('../controllers/place')
 const eventController = require('../controllers/event')
 const productController = require('../controllers/product')
-const validateJWT = require('../middleware/validateJWT')
-const validateHeaderToken = require('../middleware/validateHeaderToken')
+
+// const validateJWT = require('../middleware/validateJWT')
```
des-des: do we want to keep this? If so you should leave a comment explaining

src/routers/apiRouter.js
```diff
@@ -2,20 +2,36 @@ const userController = require('../controllers/user')
...
+// const api = require('../constants/apiPermissions')
 
 const router = require('express').Router()
 
 // user routes
 router.route('/users')
   .get(userController.getAll)
-  .post(userController.create)
+  .post(
+    // validateJWT(),
```
des-des: same as above

src/routers/apiRouter.js
```diff
@@ -2,20 +2,36 @@ const userController = require('../controllers/user')
...
+    // fieldPermissions(api.User.fields),
+    userController.create
+  )
+
 router.route('/users/:id')
   .get(userController.getById)
   .put(
-    validateJWT(),
-    validateHeaderToken,
+    // validateJWT(),
```
des-des: dito

tests/middleware/fieldPermissions.test.js
```diff
@@ -0,0 +1,267 @@
...
+  req.url = req.url.slice(5)
+  next()
+}
+
+server.post('/test/events',
+  validateHeaderToken,
+  urlSlice,
+  fieldPermissions(testPermissions),
+  (req, res, next) => res.send('success!'),
+  boomErrorHandler
```
des-des, line 89: Do you need this?

tests/middleware/fieldPermissions.test.js
```diff
@@ -0,0 +1,267 @@
...
+)
+
+server.put('/test/events/:id',
+  validateHeaderToken,
+  urlSlice,
+  fieldPermissions(testPermissions),
+  (req, res, next) => res.send('success!'),
+  boomErrorHandler
```
des-des, line 97: and this?

des-des (commented):
Leaving this halfway done, sorry will get to the other half soon!

tests/middleware/fieldPermissions.test.js
```diff
@@ -0,0 +1,267 @@
...
+    f4: 'a',
+    f5: 'a',
+    f6: 'a'
+  })
+  .expect(401)
+  .then(res => {
+    t.ok(
+      ['f1', 'f2', 'f4', 'f5'].every(field => res.body.message.includes(field)),
+      'message should list fields which are unauthorized')
+    t.notOk(
```
des-des, line 136: very nice

tests/middleware/fieldPermissions.test.js
```diff
@@ -0,0 +1,267 @@
...
+    })
+    t.end('should throw bad implementation')
+  } catch (err) {
+    t.end()
+  }
+})
+
+tape('fieldPermissions initialization', t => {
+  try {
+    fieldPermissions({
```
des-des: test with assertion can be confusing. `t.doesNotThrow`?

tests/middleware/fieldPermissions.test.js
m4v15: Yes, we are testing a fake route and some of the tests are to check we get sent back the correct error from this error handler

tests/middleware/fieldPermissions.test.js
m4v15: ☝️

src/middleware/fieldPermissions.js
m4v15: Yeah definitely we can refac this later, but I think leave this for now

mattlub: yeh- the decisions was taken because checkUserOwnsResource includes an extra db call, and isn't always necessary. But that would be cleaner for sure.

tests/middleware/fieldPermissions.test.js
mattlub: I didn't know about `t.throws` and `t.doesNotThrow`, that's what I was after.

src/routers/apiRouter.js
mattlub: it was kinda half-done already, now it's still half-done but differently.

src/routers/apiRouter.js
mattlub: I'll just delete it.

tests/middleware/fieldPermissions.test.js
mattlub: changed

m4v15:
@mattlub nice work on this, sorry I started to get this done but then other shit got in the way.

des-des:
:tada: :sparkles: good work @mattlub @m4v15

PR #152, opened 2017-09-18
Implement User permissions

m4v15:
12/10/2017

DONE! 🎉 

I've tried to make it in the same structure as @mattlub events tests using the Promise.all stuff.

- [x] I'll do the docs tomorrow morning


mattlub:
@m4v15 usage has changed, so the apiPermissions constants file needs to be updated. this should be the only thing to change though.

e.g.
```js
fields: {
  name: [ roles.SUPER, roles.OWNER]
}
```

mattlub:
merge master!

src/constants/apiPermissions.js
```diff
@@ -1,8 +1,18 @@
-// const roles = require('./roles')
+const roles = require('./roles')
 
 module.exports = {
   User: {
-
+    getAll: [roles.SUPER],
+    getById: [ roles.SUPER, roles.OWNER ],
+    create: [ roles.SUPER ],
+    update: [ roles.ADMIN, roles.OWNER ],
```
mattlub, line 8: check the [spec](https://github.com/foundersandcoders/open-tourism-platform/blob/master/data_spec_v1.md#users), it says only owner or super. What do you think?

tests/controllers/user.test.js
```diff
@@ -3,221 +3,349 @@ const supertest = require('supertest')
...
     .get('/users')
-    .expect(200)
+    .expect(401)
     .expect('Content-Type', /json/)
-    .end((err, res) => {
-      if (err) t.fail(err)
-      t.equal(res.body.length, 0, 'should initially return empty array')
-      dropCollectionAndEnd(User, t)
+    .then(res => {
+      t.equal(res.body.message, 'You are not authorised to access this resource', 'unauthorised error message sent')
```
mattlub: import the message from constants file?

tests/controllers/user.test.js
```diff
@@ -3,221 +3,349 @@ const supertest = require('supertest')
...
+    t.equal(res.body.length, 1, 'query response body should be an array with length 1')
+    t.equal(res.body[0].username, 'mattlub', 'results should be filtered correctly by url query parameters')
+    dropCollectionAndEnd(User, t)
+  })
+  .catch(err => t.end(err))
 })
 
 // GET /users/:id
-tape('GET /users/:id with valid id of something not in the database', t => {
+tape('GET /users/:id wwhen not logged in', t => {
```
mattlub: wwhen

tests/controllers/user.test.js
```diff
@@ -3,221 +3,349 @@ const supertest = require('supertest')
...
     .get('/users/507f1f77bcf86cd799439011')
-    .expect(404)
+    .expect(401)
     .expect('Content-Type', /json/)
-    .end((err, res) => {
-      if (err) t.fail(err)
-      t.equal(res.body.message, 'No document matching that id', 'response message is correct')
-      dropCollectionAndEnd(User, t)
+    .then(res => {
+      t.equal(res.body.message, 'You are not authorised to access this resource', 'unauthorised error message sent')
```
mattlub: constant message again (don't necessarily have to do it)

src/constants/apiPermissions.js
```diff
@@ -1,8 +1,18 @@
...
+    delete: [ roles.SUPER ],
+    fields: {
```
mattlub, line 10: maybe we need to add `createdAt` and `updatedAt` fields restrictions? (I didn't do this on mine either)

(Or maybe Mongoose won't allow you to edit those? But I assume it will)

tests/controllers/event.test.js
```diff
@@ -267,19 +267,6 @@ tape('DELETE /events/:id with invalid id', t => {
     })
 })
 
-tape('DELETE /users/:id with id of something not in the database', t => {
```
mattlub, line 270: please could you reinstate this, I believe it will cause a merge conflict with my branch otherwise.

mattlub (changes requested):

mattlub:
@m4v15 a damn good job, permissions on all routes...

tests/controllers/event.test.js
m4v15: @mattlub It fails the tests though and shouldn't be in there? If I reinstate it it will need to be changed anyway to pass the tests - makes more sense to fix the merge conflict if/when it happens?

src/constants/apiPermissions.js
m4v15: yeah I dunno, I think this is ok for now?

m4v15: as in, letting admin's change it is fine

src/constants/apiPermissions.js
m4v15: tbh I wrote these a few weeks ago, but I think my logic is that we probably want to allow admins to be able change the name/organisation name and stuff, but only SUPER can do username password and role - I think this is ok?

tests/controllers/user.test.js
m4v15: 👍  yeah I was being lazy, can do that now

tests/controllers/user.test.js
m4v15: 🕵️

src/constants/apiPermissions.js
mattlub: owner can change though

src/constants/apiPermissions.js
mattlub: yeh sure

tests/controllers/event.test.js
mattlub: ah ok

src/constants/apiPermissions.js
m4v15: yeah true you reckon we should add them to all the routes for super only then? it's not hard to do I guess

mattlub: ahh just leave it for now, it's not like we're using them anywhere yet.

mattlub (approved):

Issue #155, opened 2017-09-24
Shouldn't send bad implementation until the final error handler


m4v15:
relates #154 
Others it is hard to pinpoint where an error occurs - our defaults on the error switch statements should send some more information.

Issue #154, opened 2017-09-24
If location is given in the wrong format, no helpful error message returned


m4v15:
Get a `server error 500` which is no good

m4v15:
related - the location needs to come [Long, Lat] not [Lat, Long] otherwise an error happens. (Lat can be -90 to 90 and long can be -180 to 180 so in some cases if they are the wrong way round it breaks)

Should add this to the docs

m4v15:
related, we need to deal with this error in mongoErrorHandler:
```
MongoError: Can't extract geo keys
```

mattlub:
related #91 for that Mongo error

mattlub:
related #28...

Issue #91, opened 2017-07-31
POSTing invalid coordinates to /places


jwld:
Inputting coordinates outside the range results in an internal server error.

`MongoError: Can't extract geo keys`
`code: 16755`

It also allows you to input more than two coordinates in the array, but only validates the first two.

PR #157, opened 2017-09-26
Location error handler

m4v15:
addresses #154 #155 #91

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -457,7 +457,7 @@ Status: 201 Created
 Name | Type | Description
 ---|---|---
 owner | mongoose ObjectId | id of owner in user table.
-location | array of numbers | Location coordinates, [ longitude, latitude ]. [Mongo's 2dsphere Index](https://docs.mongodb.com/manual/geospatial-queries/#geospatial-legacy)
+location | array of numbers | Location coordinates, [ longitude, latitude ]. Valid longitude values are between -180 and 180, Valid latitude values are between -90 and 90 (both inclusive).
 categories | array of strings | Place [categories](#place-categories).
 accessibilityOptions | array of strings | Place [accessibility options](#accessibility-options).
 imageUrl | string | Link to image of place.
--- a/docs/api.md
+++ b/docs/api.md
@@ -371,7 +371,7 @@ Status: 200 OK
 Name | Type | Description
 ---|---|---
 owner | mongoose ObjectId | id of owner in user table.
-location | array of numbers | Location coordinates.
+location | array of numbers | Location coordinates, [ <longitude>, <latitude> ]. [Mongo's 2dsphere Index](https://docs.mongodb.com/manual/geospatial-queries/#geospatial-legacy)
 categories | array of strings | Place [categories](#place-categories).
 accessibilityOptions | array of strings | Place [accessibility options](#accessibility-options).
 imageUrl | string | Link to image of place.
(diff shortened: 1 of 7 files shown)


src/constants/errors.json
```diff
@@ -17,6 +18,7 @@
     "VALIDATION_FAILED": "Validation Failed",
     "DUPLICATE_KEY": "Data violates unique constraints validation",
     "INVALID_ID": "Invalid id",
+    "BAD_LOCATION": "The location co-ordinates do follow the correct format: [Longitude, Latitude]",
```
mattlub: missed out the word "not"

mattlub (changes requested):

src/constants/errors.json
m4v15: lol

docs/api.md
```diff
@@ -371,7 +371,7 @@ Status: 200 OK
 Name | Type | Description
 ---|---|---
 owner | mongoose ObjectId | id of owner in user table.
-location | array of numbers | Location coordinates.
+location | array of numbers | Location coordinates, [ longitude, latitude ]. [Mongo's 2dsphere Index](https://docs.mongodb.com/manual/geospatial-queries/#geospatial-legacy)
```
mattlub: I don't think the links to mongo docs are necessary/desirable, because as far as I'm concerned, a reader of the docs doesn't need to even know that we are using mongo. Thoughts? (I would also suggest changing the type of id fields to string and not Mongoose Object Id at some point)

src/constants/errors.json
```diff
@@ -17,6 +18,7 @@
     "VALIDATION_FAILED": "Validation Failed",
     "DUPLICATE_KEY": "Data violates unique constraints validation",
     "INVALID_ID": "Invalid id",
+    "BAD_LOCATION": "The location co-ordinates do not follow the correct format: [Longitude, Latitude]",
```
mattlub: add the range for each value or is it not necessary (in the docs as well maybe)?

mattlub (changes requested):

docs/api.md
m4v15: yeah actually you're right I think

mattlub (approved):

src/middleware/customErrorHandler.js
```diff
@@ -20,6 +20,6 @@ module.exports = (err, req, res, next) => {
 
     // unhandled custom error
     default:
-      res.boom.badImplementation(errMessages.UNHANDLED_CUSTOM)
```
mattlub, line 23: is  `errMessages.UNHANDLED_CUSTOM` is unused now?

src/middleware/mongoErrorHandler.js
```diff
@@ -16,9 +16,11 @@ module.exports = (err, req, res, next) => {
     case errCodes.MONGO_DUPLICATE_KEY:
       res.boom.badRequest(errMessages.DUPLICATE_KEY)
       break
-
+    case errCodes.MONGO_BAD_LOCATION:
+      res.boom.badRequest(errMessages.BAD_LOCATION)
+      break
     // unhandled mongo error
     default:
-      res.boom.badImplementation(errMessages.UNHANDLED_MONGO)
```
mattlub, line 22: same as above? @m4v15

mattlub (changes requested):

PR #159, opened 2017-10-04
fix categories that were listed incorrectly

m4v15:
Have copy and pasted from the constants file to ensure no mistakes

Any change you just wanna merge this @mattlub or @des-des as it is wrong for the data-entry and events team atm

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -917,8 +917,8 @@ Below are listed the options avaliable for each of the categories across the API
     "Disabled toilets",
     "Carer"
   ]
-    ```
-  ```json
+ ```
+ ```json
   "placeCategories": [
     "cafe",
     "music venue",
@@ -930,8 +930,8 @@ Below are listed the options avaliable for each of the categories across the API
     "municipal",
     "healthcare"
   ]
-    ```
-  ```json
+```
+```json
   "eventCategories": [
     "music",
     "dining",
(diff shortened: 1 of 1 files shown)


mattlub:
didn't we do this before? How many mistakes are there?

mattlub:
also- should we change 'miscellaneous' to 'other' in events categories? Or at least add 'other'?

m4v15:
Yeah I think we should, but not in this PR

m4v15:
Actually, none other than cultural event => culture

I guess this way was just making sure there were definitely no other mistakes.

Suuuuha:
@m4v15 @mattlub in my opinion, in the accessibility options -  'place for guide dog' could be changed to 'guide dog allowed' or sth for clarity. since in israel there are no places for guide dog but dogs might be allowed if the place is friendly accessible.

mattlub:
@m4v15 I think if it's gonna be json, we should just link to the constants file, so we don't have to change things in multiple places.

m4v15:
lol did not notice your comment here @Suuuuha. Again, we can definitely change some categories, but this PR is for fixing the currently wrong docs, **not** for changing the categories. Raise an issue?

@mattlub Yes you are probably right, I will do this now, sorry. Back to how it was originally lol.

des-des (approved):

PR #163, opened 2017-10-14
Implement Product permissions

m4v15:
- [x] add permissioning on routes
- [x] fix the tests that break because of that
- [x] add a few more permissioning tests
- [x] update the docs

### DIFF
--- a/tests/controllers/product.test.js
+++ b/tests/controllers/product.test.js
@@ -7,7 +7,7 @@ const Token = require('../../src/models/auth/Token.js')
 
 const { auth: authErrMessages } = require('../../src/constants/errors')
 
-const { validProduct1, validProduct2, invalidProduct1 } = require('../fixtures/products.json')
+const { validProduct1, validProduct2, validProduct3, invalidProduct1 } = require('../fixtures/products.json')
 const { validAdminUser, user, superUser } = require('../fixtures/users.json')
 const { token } = require('../fixtures/auth/tokens.json')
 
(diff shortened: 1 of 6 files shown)


tests/controllers/product.test.js
```diff
@@ -113,80 +123,222 @@ tape('POST /products with invalid product data', t => {
...
-              dropCollectionAndEnd(Product, t)
-            })
-            .catch(err => {
-              t.fail(err)
-              dropCollectionAndEnd(Product, t)
-            })
-        })
-    })
-    .catch(err => t.end(err))
+tape('PUT /products/:id with id of something not in the database, logged in via Oauth as SUPER', t => {
```
mattlub, line 142: oauth! :+1:

tests/controllers/product.test.js
```diff
@@ -113,80 +123,222 @@ tape('POST /products with invalid product data', t => {
...
+        dropCollectionsAndEnd([Product, User], t)
+      })
+      .catch(err => {
+        t.fail(err)
+        dropCollectionsAndEnd([Product, User], t)
+      })
+    })
+    .catch(err => t.end(err))
+  })
+  .catch(err => t.end(err))
```
mattlub: you could flatten all this to just one `catch`, it looks like all you need is the `productToBeDeleted` variable to be defined in the outermost scope

tests/index.test.js
```diff
@@ -14,6 +15,7 @@ dbConnection.once('open', () => {
     .then(() => Place.remove({}))
     .then(() => Event.remove({}))
     .then(() => Product.remove({}))
+    .then(() => Token.remove({}))
```
mattlub: Promise.all ?

mattlub (changes requested):

mattlub:
@m4v15 nice one :ear_of_rice:

tests/controllers/product.test.js
m4v15: yeah so it got complicated when accessing the id to compare to or something, I need to think it through more - I tried manually setting the id in the fixtures file but that threw a load of errors for some reason

tests/index.test.js
m4v15: yeah, this was in the middle of manic debugging so I left it for later, but yeah you're right

mattlub (approved):

des-des (approved):

Issue #162, opened 2017-10-14
check User owns resource throws 500 when resource doesn't have owner


m4v15:
Sort of an edge case but also with Data Entry it is likely a lot of resources might be made without owners initially

m4v15:
**Moved to it's own issue, left here for record sake**
Additionally, if the resource Id searched for is invalid it now gets thrown [here (when the trying to access the resource to identify the owner)](https://github.com/foundersandcoders/open-tourism-platform/blob/master/src/helpers/permissions.js#L48) so it doesn't throw an unauthorised error, even if the user is unauthorised (as this happens before it is certain if the user is authed or not).

This is not a big deal (or a deal at all?), but just got confusing with the tests for a bit

mattlub:
oops

PR #164, opened 2017-10-14
Fix checkUserOwnsResource bug

m4v15:
mentioned in this issue #162 

However, it won't fix the other bug mentioned in a comment (that bug is less critical)


mattlub:
is it worth adding a test for this?

m4v15:
hard to test before we've implemented permissions on the routes - maybe raise an issue to add tests for it after?

mattlub (approved):

PR #169, opened 2017-11-06
add mobile meta tag

m4v15:
To make login and register mobile friendly


Karyum:
# Big Changes here

Karyum:
rel #169

PR #174, opened 2018-01-17
adds event categories 

### LINKED ISSUE #173: Add more event categories
as suggested by Liwan in https://github.com/foundersandcoders/OTP-Data-Entry/issues/91 we are missing important categories Like: Movie, Theatre, Kids, Workshops.

Karyum:
closes #173

### DIFF
--- a/src/models/constants.json
+++ b/src/models/constants.json
@@ -1,9 +1,5 @@
 {
-  "roles": [
-    "BASIC",
-    "ADMIN",
-    "SUPER"
-  ],
+  "roles": ["BASIC", "ADMIN", "SUPER"],
   "accessibilityOptions": [
     "Audio recordings",
     "Braille",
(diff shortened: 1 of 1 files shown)


src/models/constants.json
```diff
@@ -39,6 +35,10 @@
     "party",
     "wedding",
     "culture",
+    "kids",
+    "movie",
```
des-des, line 39: Call this 'film' or 'movies'

des-des (changes requested):

src/models/constants.json
Karyum: yeah decided to go with `movie` instead of `film` cause `film` is more of a British way of saying it, and the society here is more american influenced :sweat_smile: , yeah but also i think being i'm picky here but `movie` makes more sense cause it's 1 event everytime and it's **a** category.

des-des (approved):

Issue #181, opened 2018-03-05
/verifed route for content verification 


Karyum:
add a GET `/verified/:id` route that will check if the user is logged in ( the user needs to be `ADMIN` or `SUPERUSER` and `OWNER` only )then will find the data and update it so verified would be true.

mattlub:
@Karyum @des-des why would verification not happen with a `PUT` request to `/resourceType/:id`?

mattlub:
Then you would just need to add the field permissions here: https://github.com/foundersandcoders/open-tourism-platform/blob/master/src/constants/apiPermissions.js

Karyum:
my mistake it's not a `GET` reuqest it is a `PUT` :sweat_smile: , but @mattlub the `/resourceType/:id` `PUT` request is also accessible to the owner of the resource, so it wouldn't be problem for a `BASIC` user to go and verify his own resource if im correct ?
 
i also didn't quite get your second comment about the field premissions?

mattlub:
@Karyum the permissions structure we built allows for different permissions on different fields, e.g. see [here](https://github.com/foundersandcoders/open-tourism-platform/blob/master/src/constants/apiPermissions.js#L37), `ADMIN` and `OWNER` can update the place but only `SUPER` can update the `id` field. 

Here's a [test](https://github.com/foundersandcoders/open-tourism-platform/blob/master/tests/controllers/place.test.js#L199-L221) for that situation.

I think a `verified` field can be implemented the same way.

Karyum:
yeah nice idea, will do that instead :+1:

Karyum:
@mattlub now thinking about it, let's say and `ADMIN` or a `SUPERUSER` is adding an event through the data entry, now you don't want them after adding the data to go to their email and verify it ( that would be ridiculous :laughing: ), you want them when they added something to the data entry it would be automatically verified cause they are not `BASIC` users, how would we go about this ?

Karyum:
hmm that was a dumb question :up: , i could just check through the `req` what is the role of the user, and toggle verified to true in the object that we are updating with.

mattlub:
cool, sorry for only noticing after you'd written all those tests 😅

PR #184, opened 2018-03-06
Limits updating verify to only ADMIN and above

Karyum:
rel #181


Karyum:
@des-des ready for review

des-des (approved):

des-des (approved):

PR #188, opened 2018-03-15
Send email

### LINKED ISSUE #187: implement send email functionality
when A `BASIC` user wants to add new content, an email is sent to all the admins, so they could verify it.

going to use [sendGrid](https://sendgrid.com/) to send emails, cause you could easily make a nice template for an email and then use something like Mustach, to dynamically change specific values.

- [x] create an account for the OTP.
- [ ] create the helper that sends emails to all admins,

Karyum:
close #187 

I have made an account on sendGrid so i would build the template on their website with HTML and CSS 
currentley doesn't look great but it has the name of the content and a link of the content in the data entry ( also arabic support :wink: ) :

*this is how literately it's gonna look like in the email* 

this is how it looks like when you recieve it :
email1

also right now everyone who is **not** a `BASIC` user is sent an email, should we keep it like that or should i make it `ADMIN` recieve emails only ?

### DIFF
--- a/src/controllers/event.js
+++ b/src/controllers/event.js
@@ -2,6 +2,7 @@ const Event = require('../models/Event')
 const { rejectIfNull } = require('../db/utils')
 const { messages: errMessages } = require('../constants/errors')
 const roles = require('../constants/roles')
+const sendEmail = require('../helpers/sendEmail')
 
 const eventController = module.exports = {}
 
@@ -50,7 +51,7 @@ eventController.create = (req, res, next) => {
 
   const newEvent = new Event(newEventDetails)
   newEvent.save()
-  // TODO: .then(event => req.user.role !== roles.BASIC ? event : sendEmail(event))
+    .then(event => req.user.role !== roles.BASIC ? event : sendEmail(event))
     .then(event => res.status(201).send(event))
     .catch(next)
 }
--- a/src/controllers/place.js
+++ b/src/controllers/place.js
@@ -2,6 +2,7 @@ const Place = require('../models/Place')
 const { rejectIfNull } = require('../db/utils')
 const { messages: errMessages } = require('../constants/errors')
 const roles = require('../constants/roles')
+const sendEmail = require('../helpers/sendEmail')
 
 const placeController = module.exports = {}
 
@@ -35,6 +36,7 @@ placeController.create = (req, res, next) => {
 
   const newPlace = new Place(newPlaceDetails)
   newPlace.save()
+    .then(place => req.user.role !== roles.BASIC ? place : sendEmail(place))
     .then(place => res.status(201).send(place))
     .catch(next)
 }
--- a/src/helpers/sendEmail.js
+++ b/src/helpers/sendEmail.js
@@ -3,36 +3,24 @@ const User = require('../models/User')
 sgMail.setApiKey(process.env.SENDGRID_API_KEY)
 sgMail.setSubstitutionWrappers('{{', '}}')
 
-const sendEmails = (emails, content) =>
-  new Promise((resolve, reject) => {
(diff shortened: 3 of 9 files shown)


mattlub:
will running the tests send real emails out? 

EDIT: I realised It will only be sending to fake addresses but still doesn't seem like a good idea

src/helpers/sendEmail.js
```diff
@@ -0,0 +1,38 @@
...
+  users.reduce((acc, curr) => {
+    if (curr.role !== 'BASIC') {
+      return acc.concat(curr.email)
+    }
+    return acc
+  }, [])
+
+module.exports = content =>
+  new Promise((resolve, reject) =>
+    User.find()
```
mattlub: you can put parameters in the `find` method to filter only the users you want, and you can also select the fields you want (just email or maybe name and email). Then you won't need the `extractEmails` call

src/helpers/sendEmail.js
```diff
@@ -0,0 +1,38 @@
...
+const extractEmails = users =>
+  users.reduce((acc, curr) => {
+    if (curr.role !== 'BASIC') {
+      return acc.concat(curr.email)
+    }
+    return acc
+  }, [])
+
+module.exports = content =>
+  new Promise((resolve, reject) =>
```
mattlub: you don't need the `new Promise(...)` part, you can do `content => User.find(...).then(...)`

src/helpers/sendEmail.js
Karyum: huh nice! didn't know this before

Karyum:
@mattlub yeah it's currently sending emails to `xxx@xxx.com` every time a test is run :sweat_smile: , not sure how to do it otherwise though :thinking:

src/controllers/event.js
```diff
@@ -50,7 +51,7 @@ eventController.create = (req, res, next) => {
 
   const newEvent = new Event(newEventDetails)
   newEvent.save()
-  // TODO: .then(event => req.user.role !== roles.BASIC ? event : sendEmail(event))
+    .then(event => req.user.role !== roles.BASIC ? event : sendEmail(event))
```
des-des, line 54: Do not put side effects in a ternary! (use an if / else)
Your async flow is cray cray, does `sendEmail` return a promise?

src/controllers/place.js
```diff
@@ -35,6 +36,7 @@ placeController.create = (req, res, next) => {
 
   const newPlace = new Place(newPlaceDetails)
   newPlace.save()
+    .then(place => req.user.role !== roles.BASIC ? place : sendEmail(place))
```
des-des, line 39: again this is silly

src/helpers/sendEmail.js
```diff
@@ -0,0 +1,26 @@
+const sgMail = require('@sendgrid/mail')
+const User = require('../models/User')
+sgMail.setApiKey(process.env.SENDGRID_API_KEY)
+sgMail.setSubstitutionWrappers('{{', '}}')
+
+const sendEmails = (emails, content) => {
+  const msg = {
+    to: emails,
+    from: 'opentourismplatform@gmail.com',
```
des-des, line 9: put this in env var?

des-des (changes requested):

src/controllers/event.js
Karyum: i don't see any problem here other than readability maybe?, if i had this in `if/else` and in ternary it's just going to do the same thing if the user is `BASIC` then sends an email then the `sendEmail` function would resolve with the event, if he's not then resolve with the event without sending the email. so whats the reason here to put this in an if statement instead of ternary ?! @des-des

